PR #1, opened 2021-12-21
Complete tests for a11y check

ethanWallace:
Use axe-core/puppeteer and E2E tests for accessibility testing.


ethanWallace:
No longer needed

PR #2, opened 2021-12-21
A11y test fail

ethanWallace:
# Purpose 
To take advantage of Stencil's spec and e2e tests when a component is generated to build accessibility testing right into the component. These tests can then be used by GitHub through actions to provide an additional layer of quality control.

This pull request was authored to show a failed colour contrast accessibility test using [axe-core/puppeteer](https://www.npmjs.com/package/@axe-core/puppeteer).

## How we can do this
Using Stencil's e2e tests, we can use the library [axe-core/puppeteer](https://www.npmjs.com/package/@axe-core/puppeteer) to run accessibility tests, this example runs a colour contrast test but can be expanded with [additional rules](https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md#wcag-21-level-a--aa-rules).

The e2e ends tests also allow testing the interactions of a component when it is loaded on a page. This example showcases a test of the `gc-ds-button` component being focused using keyboard navigation.

With a new GitHub action, when a pull request is authored/modified the tests will run for quality control.

## Modification that caused failure
The `--gcds-colour-utils-white` CSS variable has been modified from a value of `#FFFFFF` to `blue`. This causes the `gc-ds-button` to render with a text colour of `blue` and background of `#bf3230`, which causes a contrast ratio of 1.51:1 (4.5:1 required for WCAG AA)

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -1,7 +1,8 @@
 name: Run Tests
 on: 
   pull_request:
-      - main
+      branches:
+        - main
 
 jobs:
   build-deploy:
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -1,0 +1,25 @@
+name: Compile Core
+on: 
+  push:
+    branches:
+      - main
+    paths:
+      - 'src/**'
+      - 'package.json'
+
+jobs:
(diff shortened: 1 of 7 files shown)


ethanWallace:
No longer needed

PR #12, opened 2022-04-13
Site menu styles

nmakuch:
Sorry this took so long, it was slow working :s

# Summary | Résumé

Restyling the site menu component. 
The main work that was done here was:
- Replacing the background colour with something more neutral (gray here)
- Aligning the site menu container with the current container
- Adding interactive states for each menu-item (Hover, active, focus)
- Added a media query to ensure the alignment of the site-menu with the container as screen width is reduced.

## Known issues

- There is an issue with the hover state for the `components` link and the submenu trigger clashing on smaller screens (they overlap) that probably can't be resolved without changing the actual structure of the component, so leaving as is for now and we can deal with it in the future

### DIFF
--- a/src/components/gcds-site-menu/gcds-site-menu.css
+++ b/src/components/gcds-site-menu/gcds-site-menu.css
@@ -36,7 +36,7 @@
   display: inline;
 }
 :host([menu-mobile-layout*=drawer]) [data-h2-menu] {
-  background: #0a0a14;
+  background: #EDEDED;
   display: none;
   height: 100vh;
   top: 0;
@@ -165,15 +165,15 @@
   outline: none;
 }
 :host([menu-mobile-layout*=toolbar]) [data-h2-menu] {
-  background: #EDEDED;
+  background: rgba(210, 210, 210, 0.5);
 }
 @media screen and (min-width: 64em) {
   :host([menu-mobile-layout*=drawer]) [data-h2-menu] {
-    background: #EDEDED;
+    background: rgba(210, 210, 210, 0.5);
   }
 
   :host([menu-mobile-layout*=toolbar]) [data-h2-menu] {
-    background: #EDEDED;
+    background: rgba(210, 210, 210, 0.5);
   }
 }
 :host([data-h2-menu-wrapper]) [data-h2-menu].h2-mobile-menu-active {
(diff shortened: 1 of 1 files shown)


src/components/gcds-site-menu/gcds-site-menu.css
```diff
@@ -345,69 +373,147 @@
...
+:host([data-h2-menu-wrapper]) [data-h2-menu] [data-h2-menu-container] > [data-h2-menulist] > li > [data-h2-submenu-trigger]:focus span {
+  color: #ffffff !important;
+  transition: all 0.4s ease-in-out;
+}
+:host([data-h2-menu-wrapper]) [data-h2-menu] [data-h2-menu-container] > [data-h2-menulist] > li > [data-h2-submenu-trigger]:active {
+  background-color: #595959;
 }
 :host([data-h2-menu-wrapper]) [data-h2-menu] [data-h2-menu-container] > [data-h2-menulist] > li > [data-h2-menulist] {
-  background-color: #282832;
+  background-color: rgba(210, 210, 210, 0.5);
```
ethanWallace, line 412: When the looking at sub sub menu, so second level dropdown menu. The text on the page can make the bottom menu item hard to read. Maybe a slight increase in the background opacity can help with this.

src/components/gcds-site-menu/gcds-site-menu.css
```diff
@@ -651,45 +1060,52 @@
...
+  :host([menu-desktop-layout*='sidebar'][menu-alignment*='right']) [data-h2-menu] + ::slotted(div),
+  :host([menu-desktop-layout*='sidebar'][menu-alignment*='right']) [data-h2-menu] + ::slotted(main) {
     margin-left: auto;
     margin-right: 25%;
     overflow-x: hidden;
     width: 75%;
   }
 }
+
+@media screen and (max-width: 76.625em) {
```
ethanWallace, line 1107: The only time I see this take affect is when the mobile drawer is open, the menu has a gap on the right. Is this the intended style for this menu?

ethanWallace (commented):
This looks great! There were only a couple things I could see that we might need to review.

src/components/gcds-site-menu/gcds-site-menu.css
nmakuch: the site-menu sort of breaks out of the edge of the other containers without this, it only happens for a short time span, did this more or less to fix it during that tiny window

ethanWallace (approved):
LGTM

PR #13, opened 2022-04-19
Add bilingual text for required label

ethanWallace:
# Summary | Résumé

Added  passing lang attributes from gcds-input and gcds-textarea to gcds-label. When required field needs to show, it should now be bilingual.


melaniebmn (changes requested):
One comment about turning the lang function into a utility function but otherwise looks good. Thank you for catching that!

melaniebmn (approved):
Awesome, thank you for creating the util function!

PR #14, opened 2022-04-20
Fix overwriting of inherited attributes

ethanWallace:
# Summary | Résumé

Fixed issue when changing an inherited attribute would remove all other inherited attributes from shadow element. inheritedAttributes now compares itself to the shadow element as well as the host element.


melaniebmn (approved):
LGTM

PR #15, opened 2022-04-22
Update tokens

melaniebmn:
# Summary | Résumé


src/components/gcds-button/gcds-button.css
```diff
@@ -14,233 +18,137 @@
...
-      }
-      &:active, &.active {
-        
-      }
-      &:disabled, &.disabled {
-        
-      }
-    }
+  /* Shared Styles */
+  &[aria-disabled="true"] {
```
ethanWallace, line 24: Should we add a `cursor: not-allowed` to the disabled state?

ethanWallace (commented):
This is awesome! Just one small question/comment

src/components/gcds-button/gcds-button.css
melaniebmn: Good call, I just added it in.

ethanWallace (approved):

PR #20, opened 2022-05-09
Close open submenus on topbar when opening new menu

ethanWallace:
# Summary | Résumé

When opening first level submenus with the `topbar` desktop layout, opened submenus will close when opening a new submenu. 

Also added a `data-mobile` attribute to the `data-h2-menu-container` for flagging the menu is rendering mobile without relying on a specific mobile layout.


melaniebmn (approved):
Nice, LGTM

PR #21, opened 2022-05-09
Rewrite site-menu css

melaniebmn:
# Summary | Résumé

CSS rewrite for the site-menu component.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "gcds-components",
-  "version": "0.0.9",
+  "version": "0.1.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "gcds-components",
-      "version": "0.0.9",
+      "version": "0.1.0",
       "license": "MIT",
       "dependencies": {
         "@stencil/core": "^2.7.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gcds-components",
-  "version": "0.0.9",
+  "version": "0.1.0",
   "description": "gcds components proof-of-concept",
   "main": "dist/index.cjs.js",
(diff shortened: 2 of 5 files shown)


ethanWallace:
This is fantastic! Just noticed a few small things, I am going to put them all in this comment.

Not all elements with `role="menuitem"` are `<a>` tags. The menu also accepts `<span>` tags as menu items. If a span is used it currently renders like this: 

Screen Shot 2022-05-09 at 3 32 53 PM

Looks like a `align-items: center; display: flex;` could fix the issue.

---

When rendering topbar, if focused on the top-level menu item the submenu trigger button text turns white, creating a colour contrast issue.

Screen Shot 2022-05-09 at 3 27 20 PM

---

And lastly, this may come down to a token issue. I noticed the container size is really large, I believe it was this size before Nick adding styling to the site-menu. I am wondering do we adjust the tokens or do we make an easy way for someone to adjust the container width (StyleAPI)?

melaniebmn:
@ethanWallace fixed the two issues you found, thanks for pointing them out.
The container size is currently the same as the container size we are using on the docs page. I agree, it is quite big and we should give users the ability to change it. We could either do StyleAPI or we could add a `containerSize` prop that lets them select any of our supported container sizes. What do you think?

ethanWallace:
> @ethanWallace fixed the two issues you found, thanks for pointing them out. The container size is currently the same as the container size we are using on the docs page. I agree, it is quite big and we should give users the ability to change it. We could either do StyleAPI or we could add a `containerSize` prop that lets them select any of our supported container sizes. What do you think?

I think it would be best as a styleAPI property. Something like `customContainerWidth` could set the max width of the container.

melaniebmn:
@ethanWallace I'll add it as part of the sidebar changes since it's more of a feature than a rewrite.

PR #22, opened 2022-05-11
Input component updates

melaniebmn:
# Summary | Résumé

- removed red colour from the required label
- fixed input size + textarea col props to be responsive
- fix button active style


ethanWallace (approved):
LGTM!

PR #24, opened 2022-05-26
Site menu sidebar

melaniebmn:
# Summary | Résumé

Updated styling for site-menu sidebar, as well as ux for topbar.


ethanWallace:
Noticed a few styles that seemed off that should be adjusted.

1. First is the sidebar panes transparency makes the panes hard to look at.

Screen Shot 2022-05-26 at 10 52 26 AM

2. When opening a second level dropdown menu in desktop topbar and mobile drawer, the child menu becomes blue and is hard to visually see.

Screen Shot 2022-05-26 at 10 51 15 AM

ethanWallace:
There are also a few things we will need to have discussions about as a team.

With the new dropdown button/menu item layout, the accessibility keyboard controls don't flow as well as they used to so we will need to look over how we can adjust them to the new UI for the menu. Will speak with Nick and Josh sometime this sprint about how we can handle this.

Another thing we should discuss is how the sidebar layout interacts with the other site content. before you could slot the body content in to a `slot name="main"` and it would be spaced properly from the sidebar.

melaniebmn:
@ethanWallace 

> Noticed a few styles that seemed off that should be adjusted.
> 
> 1. First is the sidebar panes transparency makes the panes hard to look at.
> 
> Screen Shot 2022-05-26 at 10 52 26 AM
> 
> 2. When opening a second level dropdown menu in desktop topbar and mobile drawer, the child menu becomes blue when focused and is hard to visually see.
> 
> Screen Shot 2022-05-26 at 10 51 15 AM

It looks like you are still loading the old styles. Could you try in an incognito window or clearing your cache first? All those styling issues shouldn't be there and are not happening on my local.

Sounds good re: items to discuss.

I wasn't aware of the previous sidebar + main functionality but I can definitely add that in.

ethanWallace:
> It looks like you are still loading the old styles. Could you try in an incognito window or clearing your cache first? All those styling issues shouldn't be there and are not happening on my local.

I have tried clearing my cache and loading the menu in different browsers. the menu still displays with those styles.

ethanWallace:
This is looking great! The few things I mentioned before seem to be solved. I did notice another thing with the first menu items in topbar layout. The highest level topbar menu items with dropdown keep the old separated trigger button layout, instead of the full width trigger button.

Screen Shot 2022-06-30 at 9 33 25 AM

And one last really small detail, can we are we able to rotate the trigger button icon 90 degrees so it's pointing down when the menu is closed and up when the menu is open?

PR #25, opened 2022-06-14
Add gcds-header and gcds-footer components

ethanWallace:
# Summary | Résumé

## gcds-header

Add new `gcds-header` pattern component to the gcds. The header component will provide a a fully responsive canada.ca header with ability to customize the component using slots.

``` html
<gcds-header
  Language toggle 
  lang-href=""
 
  GC signature 
  signature-variant="colour | white"
  signature-has-link="true | false"
 
  Top navigation 
  skip-to-href=""
 
  lang="en | fr"
>
   <Slot name="topnav" />
   <Slot name="banner" />
   <Slot name="signature" />
   <Slot name="toggle" />
   <Slot name="search" />
   <Slot name="menu" />
   <Slot name="breadcrumb" />
</gcds-header>
```

### Using the component
The `gcds-header` has three built in components to interface with using attributes. Pass the values for the language toggle and top navigation components through `lang-href` and `skip-to-href` attributes.

### Customizing using slots
All elements of the `gcds-header` are able to be customized using slots. To place an element into a slot in the header, simply add `slot="{the target-slot-name}"` as an attribute to your element.

``` html
<gcds-header>
   <ul slot="menu">
   ...
   </ul>
</gcds-header>
```

## gcds-footer

Add new `gcds-footer` pattern component to the gcds. The footer component will provide a a fully responsive canada.ca footer with ability to render the full footer with the "About government" section or a compact footer with the wordmark.

``` html
<gcds-footer
   type="compact | full"
 
  GC signature 
  wordmark-variant="colour | white"
 
  Top of page link 
  top-href=""
 
  lang="en | fr"
>
   <Slot name="top" />
   <Slot name="list" />
   <Slot name="wordmark" />
</gcds-footer>
```


src/components/gcds-header/gcds-header.tsx
```diff
@@ -0,0 +1,153 @@
...
+      return;
+    }
+  }
+
+  private get renderToggle() {
+    if (!!this.el.querySelector('[slot="toggle"]')) {
+      return <slot name="toggle"></slot>;
+    } else if(this.langHref) {
+      return (
+        <section class="toggle">
```
melaniebmn, line 72: It feels weird to use a section tag for this. I know you copied this from what's on Canada.ca but maybe we can update that in the future.

melaniebmn (approved):
There are a few minor things that could be optimized, like using tokens instead of hard-coded values, but I know, you copied that code over from Canada.ca to get the exact brand values.

PR #27, opened 2022-07-11
gcds-checkbox and gcds-radio components

ethanWallace:
# Summary | Résumé

Added two new components, `gcds-checkbox` and `gcds-radio`. 

## gcds-checkbox

Renders a single checkbox with options to add additional hint or error message.

```html
<gcds-checkbox
  checkbox-id="string"
  label="string"
  name="string"

  required="boolean"
  disabled="boolean"
  value="string" 
  checked="string"
  hint="string"
  has-error="boolean"
  error-message="string"
>
</gcds-checkbox>
```

### Required fields

- `checkbox-id`
- `label`
- `name`

## gcds-radio

Renders a single radio button.

```html
<gcds-radio
  radio-id="string"
  label="string"
  name="string"

  required="boolean"
  disabled="boolean"
  value="string" 
  checked="string"
  hint="string"
  has-error="boolean"
>
</gcds-radio>
```

### Required fields

- `radio-id`
- `label`
- `name`

## Additional changes

Added `onClick` to `gcds-label` for better function with inputs, mainly the radio and checkbox.

PR #29, opened 2022-07-14
Update input and textarea components

ethanWallace:
# Summary | Résumé

Based off feedback from #27 update the internal classes of `gcds-input` and `gcds-textarea` to be name spaced classes.

## Additional changes

These have been done to avoid some potential issues with some accessibility tools.

- `aria-describedby` now only renders on components when need.
- Removed the `fieldset` in the components and replaced with `<div class="gcds-{component}-wrapper">`


melaniebmn (approved):
LGTM!

PR #32, opened 2022-07-19
Reformat site menu

ethanWallace:
# Summary | Résumé

Multiple changes to `gcds-site-menu` markup and formatting to work with newest design. 

### Dropdown button rendering

Parent item for menuitems with a dropdown now renders only as a `span` element and acts as a label for the `submenu-trigger` button. Any `a` tags passed in this spot will be converted to a `span`.

### Sidebar back buttons + backdrop

Sidebar layout now has additional back buttons included in each submenu pane at the top of the menulist. 

When one of the submenus is active, a backdrop will cover the rest of the page content to prevent clicking outside elements. Will close when the submenus are closed or the backdrop is clicked.

### Conditional render slots

The `left` and `right` slots will only render if something is passed to them. This should give more space for menu items in the `topbar` layout.

### Topbar focus out

If focus leaves the `gcds-site-menu` when submenus are open in the `topbar` layout, close the submenus.

### Responsive main slot with sidebar

Passing content to the `slot="main"` while using the `sidebar` layout will place the content in a responsive area next to the sidebar to prevent overlap.

### Updated tests

Updated tests so they no longer fail. Also added some tests to cover more layout options.


melaniebmn (approved):
Nice work! :) Let's merge this and hopefully we won't have to work on this component for a while haha

PR #39, opened 2022-08-18
add gcds-stepper component

melaniebmn:
# Summary | Résumé

Adding gcds-stepper component - A stepper guides users through multi-step flows with numbered steps.

Properties:
- **current-step**: The step the user is currently on
- **total-steps**: The total number of steps
- **lang**: Language of the component

### DIFF
--- a/src/components/gcds-stepper/gcds-stepper.tsx
+++ b/src/components/gcds-stepper/gcds-stepper.tsx
@@ -1,0 +1,44 @@
+import { Component, Element, Host, Prop, h } from '@stencil/core';
+import { assignLanguage } from '../../utils/utils';
+
+@Component({
+  tag: 'gcds-stepper',
+  styleUrl: 'gcds-stepper.css',
+  shadow: true,
+})
+export class GcdsStepper {
+  private lang: string;
+
+  @Element() el: HTMLElement;
+
+  /**
+   * Defines the current step.
+   */
+  @Prop() currentStep: number;
+
+  /**
+   * Defines the total amount of steps.
+   */
+  @Prop() totalSteps: number;
+
+  async componentWillLoad() {
+    // Define lang attribute
+    this.lang = assignLanguage(this.el);
+  }
+
+  render() {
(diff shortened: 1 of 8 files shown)


src/components/gcds-stepper/gcds-stepper.tsx
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  render() {
+    const { currentStep, lang, totalSteps } = this;
+
+    return (
+      <Host>
+        <nav class="gcds-stepper" role="navigation">
+          { lang == 'en' ?
+            <h6>Step {currentStep} of {totalSteps}</h6>
```
ethanWallace, line 36: To add emphasis to the stepper text it might be better to use a `<strong>` tag. Using an `<h6>` will probably create issues with jumping heading levels.

ethanWallace (commented):
Just a few things I noticed

src/components/gcds-stepper/gcds-stepper.tsx
melaniebmn: I chose the `<h6>` because I think it is partially a navigational element as it navigates the user through a form and tells them what step they are currently on. Having it added to the heading levels could be beneficial for screen reader users because I think only using `<strong>` doesn't give it enough importance. What do you think?

src/components/gcds-stepper/gcds-stepper.tsx
```diff
@@ -0,0 +1,44 @@
...
+    // Define lang attribute
+    this.lang = assignLanguage(this.el);
+  }
+
+  render() {
+    const { currentStep, lang, totalSteps } = this;
+
+    return (
+      <Host>
+        <nav class="gcds-stepper" role="navigation">
```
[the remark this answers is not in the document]
melaniebmn: Yeah fair enough, I was thinking of it as a partial navigation element but `<nav>` should really only be used if there are links included. I updated it.

ethanWallace (approved):
LGTM

PR #41, opened 2022-08-25
add gcds-alert component

melaniebmn:
# Summary | Résumé

Add gcds-alert component.

Alerts notify users of important and often time-sensitive changes. They are usually positioned at the top of the page to capture the user’s attention and therefore should be concise and used sparingly.

- **alert-heading**: Text for the alert heading
- **alert-role**: Defines the role of the alert
- **hide-close-btn**: Controls if the close button is displayed or not
- **onDismiss**: Callback when the close button is clicked
- **max-content-width**: Defines the container width for the alert content
- **position-fixed**: Controls if the alert is fixed to the top or not

Update: The french translations have been added.

### DIFF
--- a/src/components/gcds-alert/gcds-alert.tsx
+++ b/src/components/gcds-alert/gcds-alert.tsx
@@ -25,7 +25,7 @@ export class GcdsAlert {
   /**
    * Callback when the close button is clicked.
    */
-  @Prop() dismissHandler: () => void;
+  @Prop() dismissHandler: Function;
 
   /**
    * Defines if the alert's close button is displayed or not.
@@ -59,11 +59,11 @@ export class GcdsAlert {
 
   @Event() gcdsDismiss!: EventEmitter<void>;
 
-  private onDismiss = () => {
+  private onDismiss = (e) => {
     this.gcdsDismiss.emit();
 
     if ( this.dismissHandler ) {
-      this.dismissHandler();
+      this.dismissHandler(e);
     } else {
       this.isOpen = false;
     }
@@ -114,7 +114,7 @@ export class GcdsAlert {
                 { !hideCloseBtn ?
                   <button
                     class="alert-close-btn"
-                    onClick={this.onDismiss}
+                    onClick={(e) => this.onDismiss(e)}
                     aria-label={ lang == 'en' ? 'Close alert.' : 'Fermer l\'alerte.'}
                   >
                     <gcds-icon aria-hidden="true" name="times" size="sm" />
--- a/src/components/gcds-alert/gcds-alert.tsx
+++ b/src/components/gcds-alert/gcds-alert.tsx
@@ -36,15 +36,15 @@ export class GcdsAlert {
    */
   @Prop() positionFixed?: boolean = true;
 
-
-  /**
-  * Events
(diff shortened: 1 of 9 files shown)


src/components/gcds-alert/gcds-alert.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * Defines if the alert's position is fixed.
+   */
+  @Prop() positionFixed?: boolean = true;
+
+  /**
+   * Callback when the close button is clicked.
+   */
+  @Prop() onDismiss: () => void;
```
ethanWallace, line 42: To avoid having build warning everywhere for when we add this functionality to other components, maybe we should switch the naming of these types of props to `{event}Handler` instead of `on{event}`, so `onDismiss` would become `dismissHandler`.

src/components/gcds-alert/gcds-alert.tsx
```diff
@@ -0,0 +1,108 @@
...
+    const { alertHeading, alertRole, hideCloseBtn, lang, maxContentWidth, onDismiss, positionFixed } = this;
+
+    return (
+      <Host>
+        <div
+          class={`gcds-alert role-${alertRole} ${positionFixed ? 'is-fixed' : ''}`}
+          role="alert"
+          aria-label={
+            lang == 'en' ?
+              `This is ${ alertRole == 'info' ? 'an' : 'a'} ${alertRole} alert.`
```
ethanWallace, line 68: I really like this label but I am worried that when using the alert in `alert-role="destructive"` that the term destructive won't be clear to the user. It is consistent with our naming scheme but that's something a developer really only sees.

Doesn't mean it has to change, just want to flag it for when we do testing.

src/components/gcds-alert/gcds-alert.tsx
```diff
@@ -0,0 +1,108 @@
...
+                : alertRole === 'warning' ? 'exclamation-triangle'
+                : null }
+              />
+
+              <span>{alertHeading}</span>
+
+              { !hideCloseBtn ?
+                <button
+                  class="alert-close-btn"
+                  onClick={onDismiss ? onDismiss : null}
```
ethanWallace, line 92: We should also include a function built in to dismiss the alert as well. From my research so far, it looks like passing a function is a very React way to handle these types of situations so we should have a way to control it for other frameworks or teams not using a framework.

ethanWallace (changes requested):
Just noticed a few things to discuss but looking good otherwise!

src/components/gcds-alert/gcds-alert.tsx
melaniebmn: That's a good point, I'm going to change it to 'critical' for the label instead. I will also add a note for myself to discuss with Amy if we should add any additional content for the destructive alert specifically, for example something like "Please take immediate reaction".

src/components/gcds-alert/gcds-alert.tsx
melaniebmn: @ethanWallace updated!

ethanWallace (approved):
LGTM!

PR #43, opened 2022-09-01
add gcds-details component

melaniebmn:
# Summary | Résumé

Adding the gcds-details component.

The details component (also called accordion) is an interactive title/summary that shows or hides related, additional content when selected. It is typically used in a list and provides the user with a high-level overview of content that is not crucial to read in full. It reduces the need to scroll.

Properties
details-title: The title for the details component summarizes the panel content.
open: Defines if the details panel is open by default or not.

Slotted Elements

Slot
The details slot is content that is associated with the details title. It can include text only or additional elements such as buttons that allow the user to take action.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gcds-components",
-  "version": "0.1.8",
+  "version": "0.1.9",
   "description": "gcds components proof-of-concept",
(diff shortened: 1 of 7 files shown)


ethanWallace (approved):
LGTM!

PR #47, opened 2022-09-15
Add extra props + custom callback functions to components

ethanWallace:
# Summary | Résumé

Form components can now accept custom callback function passed through the props `focusHandler`, `blurHandler`, `changeHandler` and `clickHandler`. This should increase flexibility working with react.

Form compoennts can now accept additional `aria-describedby` values, these will be added to the default list created by the component.


melaniebmn (approved):
LGTM

PR #57, opened 2022-10-27
Change gcds-footer type prop to display

ethanWallace:
# Summary | Résumé

Change `gcds-footer`'s  `type` prop to `display` to better reflect what the property does.


melaniebmn (approved):
LGTM

PR #74, opened 2022-12-09
Props and classes update

ethanWallace:
# Summary | Résumé

Update props and classes to align naming conventions.


packages/web/src/components/gcds-radio/test/gcds-radio.spec.tsx
```diff
@@ -72,7 +72,7 @@ describe('gcds-radio', () => {
     });
     expect(page.root).toEqualHtml(`
       <gcds-radio radio-id="radio" disabled="" label="radio" name="radio">
-        <div class="gcds-radio-wrapper gcds-disabled">
+        <div class="gcds-disabled gcds-radio">
```
melaniebmn, line 75: Should this be `gcds-radio--disabled` if we are using `gcds-radio--error`?

melaniebmn (approved):
LGTM

packages/web/src/components/gcds-radio/test/gcds-radio.spec.tsx
ethanWallace: Good catch! Was updated in the CSS just missed it on the component/test.

PR #78, opened 2023-01-04
Update gcds footer with new design

ethanWallace:
# Summary | Résumé

Update `gcds-footer` to match [new design pattern](https://design.canada.ca/common-design-patterns/site-footer.html).

### Changes
- Removed `topHref` property
- Changed links in main section
- Add new contextual section

#### Using contextual section

The contextual section will render after using the `contextualHeading` and `contextualLinks` props.

- `contextualHeading` expects a string.
- `contextualLinks` expects a JSON object in a string for HTML or an object when using a jsx environment 

```
    <gcds-footer
      display="compact"
      contextual-heading="Heading"
      contextual-links='{ "link 1": "#red","link 2": "#green", "link 3": "#blue", "link4": "#yellow" }'
    >
    </gcds-footer>
    ```

### CODE REVIEW

File: packages/web/src/components/gcds-footer/gcds-footer.tsx [hunk 1]
```diff
@@ -66,29 +61,51 @@ export class GcdsFooter {
...
     return (
       <Host role="contentinfo">
         <div class="top__container">
           <slot name="top"></slot>
         </div>
-
+        {hasContextual && 
+          <div class="gcds-footer__contextual">
+            <div class="contextual__container">
+              <slot name="contextual"></slot>
```
melaniebmn, line 82: Instead of adding a slot and trusting the user to add the proper tags (aka nav + h3 heading, ul), should we include these and only use the slot to let users add their own list items? It would ensure better accessibility since we could set up the heading as nav label etc. We could add a prop for the contextual heading?

packages/web/src/components/gcds-footer/gcds-footer.tsx
```diff
@@ -66,29 +61,51 @@ export class GcdsFooter {
...
-    const { lang, display, topHref, hasList, renderSignature } = this;
+    const { lang, display, hasList, hasContextual, renderSignature } = this;
     const govNav = I18N[lang].gov.menu;
+    const themeNav = I18N[lang].themes.menu;
     const siteNav = I18N[lang].site.menu;
 
     return (
       <Host role="contentinfo">
         <div class="top__container">
           <slot name="top"></slot>
```
melaniebmn, line 77: Should we remove this slot as it is most likely no longer needed?

packages/web/src/components/gcds-footer/gcds-footer.tsx
```diff
@@ -66,29 +61,51 @@ export class GcdsFooter {
...
+                  {Object.keys(themeNav).map((value) =>
+                    <li>
+                      <a href={themeNav[value].link}>{themeNav[value].text}</a>
+                    </li>
+                  )}
+                </ul>
+              </nav>
+            </div>
           </div>)
         : null }
```
melaniebmn, line 111: Looks like the footer brand is also using an `<h3>` now instead of the `<h2>`. They also updated the footer classes. The brand footer is now "gc-sub-footer" and the landscape footer is now "gc-main-footer".

packages/web/src/components/gcds-footer/gcds-footer.tsx
```diff
@@ -66,29 +61,51 @@ export class GcdsFooter {
...
+                <h3>{I18N[lang].gov.heading}</h3>
+                <ul>
+                  {Object.keys(govNav).map((value) =>
+                    <li>
+                      <a href={govNav[value].link}>{govNav[value].text}</a>
+                    </li>
+                  )}
+                </ul>
+              </nav>
+              <nav class="landscape__themenav" aria-label={I18N[lang].themes.heading}>
```
melaniebmn, line 98: In the example footer, both of these lists are in the same nav element. Did you split them up into 2 nav's to have them both show up as landmarks?

packages/web/src/components/gcds-footer/gcds-footer.css
```diff
@@ -1,18 +1,19 @@
...
-  h2 {
-    display: none;
-  }
-
-  a {
-    font-size: var(--gcds-font-sizes-caption);
+  font-size: 0.875rem;
+
+  h2,
+  h4 {
```
melaniebmn, line 7: Maybe use a class for this instead to avoid accidentally hiding headings?

packages/web/src/components/gcds-footer/gcds-footer.css
```diff
@@ -139,44 +170,38 @@
...
+        }
+
         @media screen and (min-width: 64em) {
           display: inline-block;
         }
 
         &:before {
-          content: "\2022";
-          margin-right: var(--gcds-spacing-300);
+          display: none;
```
melaniebmn, line 191: Since you are only styling the before pseudo-class within the media query now, you should be able to remove this.

melaniebmn (changes requested):
Added some comments/items to discuss. Mainly, I think the contextual slot should include the nav, h3, ul and only allow users to add their own list items within the slot to ensure the correct elements are present and that the nav element is used together with a proper heading.

melaniebmn:
Oops and one last item, when I added the contextual footer, it didn't seem to load the correct styling:

Screenshot 2023-01-05 at 10 00 44 AM

packages/web/src/components/gcds-footer/gcds-footer.tsx
ethanWallace: Yes I decided to split the larger nav element into two different nav elements, since the first links are more global in context and the themes and topics are more specific.

Do you think I should format it the way canada.ca does or do you think this way will also work?

packages/web/src/components/gcds-footer/gcds-footer.tsx
ethanWallace: Agreed, will remove top slot.

ethanWallace:
> Oops and one last item, when I added the contextual footer, it didn't seem to load the correct styling:

That's unfortunately one of the downside of slots. You are only able to style one level down in slotted content, so if you pass a `nav`, `ul`, `li` and `a` tags into the slot. You would only be able to apply styling directly to the `nav` element. Developers using the contextual slot will have to style that menu themselves.

melaniebmn (changes requested):
Footer is looking good overall, my only suggestion: instead of using a slot for the contextual menu, let users pass in a link object that then displays the list so that the user doesn't have to worry about styling.

PR #80, opened 2023-01-09
Fix/gcds details panel

amazingphilippe:
# Summary | Résumé

Just a quick fix to target `details__panel` instead of `details-panel`


melaniebmn (approved):
LGTM

PR #81, opened 2023-01-10
improve breadcrumbs component

melaniebmn:
# Summary | Résumé

Improvements to breadcrumbs component:
- Nicholas recommended using `chemin de navigation` for breadcrumbs instead of `fil d'Ariane`. We'll include fil d'Ariane as an option for the `other names` section.
- Removed current page to avoid confusion with users.

### DIFF
--- a/packages/angular/src/lib/stencil-generated/components.ts
+++ b/packages/angular/src/lib/stencil-generated/components.ts
@@ -40,13 +40,13 @@ export declare interface GcdsBreadcrumbs extends Components.GcdsBreadcrumbs {}
 
 @ProxyCmp({
   defineCustomElementFn: undefined,
-  inputs: ['currentPageTitle', 'hideCanadaLink']
+  inputs: ['hideCanadaLink']
 })
 @Component({
   selector: 'gcds-breadcrumbs',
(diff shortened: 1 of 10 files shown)


ethanWallace (commented):
Instead of assigning the current page from the top level breadcrumbs component, is there anyway we can move the logic of assigning `aria-current="page"` down to `gcds-breadcrumbs-item`?

That way the user can still pass an href to the current page breadcrumb item so it will be a focusable item.

melaniebmn:
Yes, that's possible and it is what I had before but I noticed in the design, that the current page is supposed to not be a link:

https://www.figma.com/file/mh2maMG2NBtk41k1O1UGHV/GC-Components-Main?node-id=2353%3A7841&t=WzsS7eV71WM63KvH-0

I also read in a few other design systems that the current page shouldn't be selectable. The W3 pattern does have the current page as a link.

Do we want to follow the W3 again?

ethanWallace:
Either way works, as long as we can provide reasoning as to why we do it the current way it is.

ethanWallace (approved):
Since the design matches what is currently in our Figma, it looks good!

PR #85, opened 2023-01-26
Varous component and token fixes

ethanWallace:
# Summary | Résumé

Update component tokens to fix display issues.

- `gcds-site-menu`: Change container token to current token
- `gcds-header`: Add `!important` to avoid overwriting margin styles
- `gcds-breadcrumbs`:Add `!important` to avoid overwriting margin styles on slotted first-child list item
- `gcds-footer`: Token update to fix wordmark margin

### DIFF
--- a/packages/web/src/components/gcds-site-menu/gcds-site-menu.css
+++ b/packages/web/src/components/gcds-site-menu/gcds-site-menu.css
@@ -426,7 +426,7 @@
 
     [data-h2-menu-container] {
       display: flex;
-      max-width: var(--gcds-container-xl);
+      max-width: var(--gcds-site-menu-container-max-width);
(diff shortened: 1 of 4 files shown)


packages/web/src/components/gcds-site-menu/gcds-site-menu.css
```diff
@@ -426,7 +426,7 @@
 
     [data-h2-menu-container] {
       display: flex;
-      max-width: var(--gcds-container-400);
+      max-width: var(--gcds-container-xl);
```
melaniebmn, line 429: Can you add a component token for this? Thank you!!

melaniebmn (approved):
One small request, but otherwise LGTM

PR #86, opened 2023-01-26
update alert design + tokens

melaniebmn:
# Summary | Résumé

Update alert design + add new tokens.
- Added a new prop to hide the role icon.
- Renamed "max-content-width" prop to "container" for better consistency with other components.

### DIFF
--- a/packages/angular/src/lib/stencil-generated/components.ts
+++ b/packages/angular/src/lib/stencil-generated/components.ts
@@ -18,13 +18,13 @@ export declare interface GcdsAlert extends Components.GcdsAlert {
 
 @ProxyCmp({
   defineCustomElementFn: undefined,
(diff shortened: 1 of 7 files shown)


amazingphilippe:
>  Why do we not use our usual outline focus

We can definitely use our usual outline focus. Good point. Looking at this, why don't we also give a bit more contrast to those buttons either on hover or on default?

image

amazingphilippe:
Another option: 
image

With prototype: https://www.figma.com/proto/mh2maMG2NBtk41k1O1UGHV/GC-Components-Main?page-id=885%3A2249&node-id=2928%3A11518&viewport=590%2C-140%2C0.39&scaling=min-zoom

ethanWallace:
I meant a focus state more like the button we have, just to keep the focus states consistent when we use buttons. See below:

Screen Shot 2023-01-26 at 4 22 24 PM

I don't think every other state requires a redesign

amazingphilippe:
Oh I see what you mean. We just missed the extra border: 
image

PR #91, opened 2023-02-01
Icon font size

amazingphilippe:
# Summary | Résumé

## Updates to the icon component
- Sizes match with font size
- Supports a fixed width padding prop to create square icons

## Updates to components using `<gcds-icon>`
- File uploader hover and active states icon colors now work
- The icon is contained in a box, as designed in Figma
- The alert icon uses the h5 size
- Removed conflicting font declarations on`.alert__icon`

This PR also needs a token update to work properly. https://github.com/cds-snc/gcds-tokens/pull/84


packages/web/src/components/gcds-file-uploader/gcds-file-uploader.css
```diff
@@ -128,16 +124,25 @@
...
     pointer-events: none;
-    color: var(--gcds-file-uploader-file-icon-default);
+    color: var(--gcds-file-uploader-file-icon-text);
+    background-color: var(--gcds-file-uploader-file-icon-background-default);
     transition: color .15s ease-in-out;
+    border-radius: var(--gcds-file-uploader-file-border-radius);
+
+    /* Centering icon in a min-target box */
+    width: 44px;
+    height: 44px;
```
amazingphilippe, line 134: Maybe those could be tokens.

packages/web/src/index.html
```diff
@@ -4,33 +4,37 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=5.0" />
     <title>Stencil Component Starter</title>
+
+    <!-- Utility framework -->
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/cds-snc/gcds-utilities/dist/utilities.css" />
+
+
```
amazingphilippe, line 11: Importing first so that this demo page can overwrite utilities. For example, font awesome would load from utilities, instead of from this index. Usually isn't it better to place utilities first?

packages/web/src/index.html
```diff
@@ -4,33 +4,37 @@
...
 
-    <!-- Utility framework -->
-    <link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/cds-snc/gcds-utilities/dist/utilities.css" />
+    <!-- Font awesome 6 kit for this demo page. Anyone can create a free kit for open source. -->
+    <script src="https://kit.fontawesome.com/892cb27850.js" crossorigin="anonymous"></script>
+
+    <!-- Local components -->
+    <link rel="stylesheet" href="/dist/gcds/gcds.css" />
+    <script type="module" src="/dist/gcds/gcds.esm.js"></script>
+    <script nomodule src="/dist/gcds.js"></script>
```
amazingphilippe, line 28: Grouping all fonts above, then font awesome kit, then local stuff

packages/web/src/components.d.ts
```diff
@@ -390,26 +390,34 @@ export namespace Components {
         "hintId": string;
     }
     interface GcdsIcon {
+        /**
+          * Style of the icon.
+         */
+        "iconStyle"?: 'regular' | 'solid';
```
amazingphilippe, line 396: Regular is outlined icons
Solid is filled

amazingphilippe: The solid style has most icons. Regular is mostly outlined variations for solid icons.

melaniebmn: What about using `outlined` instead of `regular`? Makes it a little bit easier to understand the difference with `solid` icons.

packages/web/src/components/gcds-icon/gcds-icon.css
```diff
@@ -1,4 +1,4 @@
-@import url('https://unpkg.com/font-awesome/css/font-awesome.min.css');
+@import url('font-awesome.min.css');
```
melaniebmn, line 1: The import was set up to always load the newest version of font awesome so that we don't have to manually update it ourselves. Do you have any concerns about doing it that way or why did you change it?

packages/web/src/components/gcds-icon/gcds-icon.tsx
```diff
@@ -36,19 +40,25 @@ export class GcdsIcon {
   /**
    * Defines the size of the icon.
    */
-  @Prop() size?: 'inherit' | 'xs' | 'sm' | 'md' | 'lg' | 'xl' = 'inherit';
+  @Prop() padding?: 'none' | 'fixed-width' = 'none';
```
melaniebmn, line 43: Since this prop doesn't actually add any padding, but calculates the width instead, how about renaming this prop to `fixedWidth` and making it a boolean? If it's true, you can add the `fixed-width` class that calculates the fixed with, otherwise, the class won't be added.

packages/web/src/components/gcds-icon/gcds-icon.css
```diff
@@ -136,29 +136,81 @@
     margin-inline-end: var(--gcds-icon-margin-1000);
   }
 
+  /* Padding */
+  &.padding-fixed-width {
+    text-align: center;
+  }
+
+  &.size-caption.padding-fixed-width {
+    width: calc(var(--gcds-icon-fixed-width-caption) * 1em);
```
melaniebmn, line 145: We might have to go a bit bigger than 1em.
When I tested the fixed width with the cloud icon, the icon was bigger than the width:

Screenshot 2023-02-06 at 11 13 18 AM

melaniebmn (changes requested):
A few comments/suggestions, but mostly good to go!

packages/web/src/components/gcds-icon/gcds-icon.css
amazingphilippe: Yes1 The latest unpkg font awesome version is not up to date anymore. Maybe we should consider the npm package instead.

packages/web/src/components/gcds-icon/gcds-icon.tsx
amazingphilippe: yeah, lets do that!

packages/web/src/components/gcds-icon/gcds-icon.css
amazingphilippe: I think this is on purpose from font-awesome. The cloud might be bigger for optical reasons. Similar to baseline overshoot in typography: 
image

packages/web/src/components.d.ts
melaniebmn: Good point, in that case, could you please add a simple description to the prop comment?
Something very easy like "the regular style, will use the outlined version of the icon" or something like that.

packages/web/src/components/gcds-icon/gcds-icon.css
melaniebmn: Oh weird, unpkg should always have the latest version.
Let's use the npm package instead then.

packages/web/src/components/gcds-icon/gcds-icon.css
melaniebmn: Yeah could be, ok I'll resolve the comment for now. Maybe some we can research.

packages/web/package-lock.json
```diff
@@ -1,11396 +1,8 @@
 {
   "name": "gcds-components",
   "version": "0.1.15",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
melaniebmn, line 4: How come all packages are being removed from here? Was that by mistake?

packages/web/src/components.d.ts
```diff
@@ -1437,26 +1445,34 @@ declare namespace LocalJSX {
         "hintId"?: string;
     }
     interface GcdsIcon {
+        /**
+          * Defines the size of the icon.
+         */
+        "fixedWidth"?: boolean;
```
melaniebmn, line 1451: Can you set the default to `false` please?

packages/web/package-lock.json
amazingphilippe: Probably, what does that do? I installed the font awesome package. That's the only thing I can see.

packages/web/src/components/gcds-file-uploader/gcds-file-uploader.tsx
```diff
@@ -272,55 +272,47 @@ export class GcdsFileUploader {
...
-              onBlur={(e) => this.onBlur(e)}
-              onFocus={(e) => this.onFocus(e)}
-              onChange={(e) => this.handleChange(e)}
+              onBlur={e => this.onBlur(e)}
+              onFocus={e => this.onFocus(e)}
+              onChange={e => this.handleChange(e)}
               aria-invalid={hasError ? 'true' : 'false'}
             />
-            { value.length > 0 ?
+            {value.length > 0 ? (
```
melaniebmn, line 296: You can omit the brackets here, there is no need for them.

packages/web/src/components/gcds-file-uploader/gcds-file-uploader.tsx
```diff
@@ -272,55 +272,47 @@ export class GcdsFileUploader {
...
-            <button
-              class="file-uploader__uploaded-file"
-              onClick={(e) => this.removeFile(e)}
-              aria-label={ lang == 'en' ? `Remove file ${file}` : `Supprimer le fichier ${file}` }
-            >
-              <span>{file}</span>
-              <gcds-icon name="times-circle" size="lg" />
-            </button>
-          )) : null }
+          {value.length > 0
```
melaniebmn, line 308: Please revert this back to being on one line, makes it easier to read or at least, leave the `?` in this line.

packages/web/src/components/gcds-file-uploader/gcds-file-uploader.tsx
```diff
@@ -272,55 +272,47 @@ export class GcdsFileUploader {
...
+            ? value.map(file => (
+                <button class="file-uploader__uploaded-file" onClick={e => this.removeFile(e)} aria-label={lang == 'en' ? `Remove file ${file}` : `Supprimer le fichier ${file}`}>
```
melaniebmn, line 310: Please revert this back to how it was formatted before. Ethan and I purposely separate each prop on a separate line when they overflow the line to make it easier to read,

packages/web/src/components/gcds-icon/gcds-icon.css
```diff
@@ -136,29 +136,81 @@
     margin-inline-end: var(--gcds-icon-margin-1000);
   }
 
+  /* Padding */
```
melaniebmn, line 139: Can you update this comment to say `Fixed width` instead of `padding` please since the prop changed?

packages/web/src/components/gcds-icon/gcds-icon.tsx
```diff
@@ -21,34 +25,40 @@ export class GcdsIcon {
...
 
   /**
    * Name of the icon.
    */
   @Prop() name!: string;
 
+  /**
+   * If the icon should render as a fixed-width square, or their natural width.
+   */
+  @Prop() fixedWidth: boolean;
```
melaniebmn, line 43: Can you set the default value to `false` here?

melaniebmn (approved):
Added a few more small comments.

PR #92, opened 2023-02-01
A11y updates for components

ethanWallace:
# Summary | Résumé

Accessibility fixes for various components. 

## Broken aria reference

Modified the way `aria-describedby` is formatted in form components to remove extra whitespace between ids. WAVE toolbar would identify these as errors.

## Breadcrumb list

Modified `gcds-breadcrumbs-item` to remove the `li` and assigned `role="listitem"` to the host element to fix an issue with lists being strucutred properly. Flagged using Accessibility Insights for Web.


packages/web/src/components/gcds-breadcrumbs/gcds-breadcrumbs-item.tsx
```diff
@@ -22,12 +22,10 @@ export class GcdsBreadcrumbsItem {
     const { href } = this;
 
     return (
-      <Host>
-        <li class="gcds-breadcrumbs-item">
-          <a href={href}>
-            <slot></slot>
-          </a>
-        </li>
+      <Host role="listitem">
```
melaniebmn, line 25: Can we keep the class here and add it to the host instead?

packages/web/src/components/gcds-breadcrumbs/gcds-breadcrumbs-item.css
```diff
@@ -1,8 +1,7 @@
-:host .gcds-breadcrumbs-item {
+:host {
```
melaniebmn, line 1: Same comment about the class

packages/web/src/components/gcds-breadcrumbs/gcds-breadcrumbs-item.css
```diff
@@ -1,8 +1,7 @@
...
   position: relative;
   display: inline-block;
-  margin: var(--gcds-breadcrumbs-item-margin);
-  padding: var(--gcds-breadcrumbs-item-padding);
+  padding: var(--gcds-breadcrumbs-item-padding) !important;
```
melaniebmn, line 4: We can't remove the margin from the breadcrumb item, it's to create some spacing between the breadcrumb items on smaller screens.

packages/web/src/components/gcds-breadcrumbs/gcds-breadcrumbs.css
```diff
@@ -4,7 +4,7 @@
   ol {
     font: var(--gcds-breadcrumbs-font);
     list-style: none;
-    margin: 0;
+    margin: var(--gcds-breadcrumbs-margin);
```
melaniebmn, line 7: Same comment about removing the margin.

melaniebmn (changes requested):
One comment about removing the margin from the breadcrumb items, everything else looks good.

packages/web/src/components/gcds-breadcrumbs/gcds-breadcrumbs-item.css
ethanWallace: The bread crumb items are having some conflicts with ```* {
    -webkit-box-sizing: border-box;
    box-sizing: border-box;
    margin: 0;
    padding: 0;
}```
after the latest change. I am going to try and move the margin to the `a` tags instead to keep the spacing for mobile.

ethanWallace:
@melaniebmn I moved the margin back into the item but onto the a tag now. The breadcrumb item has a bigger focus state now but I kind of like the new look of the focus state.

melaniebmn (approved):
LGTM

PR #95, opened 2023-02-07
Add new footer prop: sub-links

ethanWallace:
# Summary | Résumé

Add new prop `sub-links` to `gcds-footer` and removed `list` slot.


melaniebmn (approved):
Nice, looks great! The only question I have is: do we need to update any of the test cases for this change?

ethanWallace:
Good point! I will add additional tests in to make sure this is covered.

PR #97, opened 2023-02-07
Icon font size

amazingphilippe:
# Summary | Résumé

## Updates to the icon component
- Sizes match with font size
- Supports a fixed width padding prop to create square icons

## Updates to components using `<gcds-icon>`
- File uploader hover and active states icon colors now work
- The icon is contained in a box, as designed in Figma
- The alert icon uses the h5 size
- Removed conflicting font declarations on`.alert__icon`


melaniebmn (approved):
LGTM

PR #107, opened 2023-02-22
Update actions to install correct packages

ethanWallace:
# Summary | Résumé

Update github actions to install only web package to run tests.


melaniebmn (approved):
LGTM

PR #111, opened 2023-02-28
Add French side of READMEs for all packages

ethanWallace:
# Summary | Résumé

Add French translations for all package READMEs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-[La version française suit.](#système-de-design---composants-gc)
+[La version française suit.](#composants-de-système-de-design-gc)
 
 # GC Design System Components
 
@@ -22,7 +22,7 @@ You can find the full documentation for GC Design System Components on [https://
 
 - Clone the repo git clone `https://github.com/cds-snc/gcds-components`.
 - Run `npm install` to install all Node.js dependencies.
-- Run `npm install –workspaces` to install all Node.js dependencies for the web, React and Angular packages.
+- Run `npm install -–workspaces` to install all Node.js dependencies for the web, React and Angular packages.
 - Change into the `/packages/web` directory.
 - Run `npm run build` to compile web components.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -22,7 +22,7 @@ You can find the full documentation for GC Design System Components on [https://
 
 - Clone the repo git clone `https://github.com/cds-snc/gcds-components`.
```
melaniebmn, line 23: Can you include `git clone` in the command as well pls?

README.md
```diff
@@ -22,7 +22,7 @@ You can find the full documentation for GC Design System Components on [https://
...
 - Run `npm install` to install all Node.js dependencies.
-- Run `npm install –workspaces` to install all Node.js dependencies for the web, React and Angular packages.
+- Run `npm install -–workspaces` to install all Node.js dependencies for the web, React and Angular packages.
 - Change into the `/packages/web` directory.
 - Run `npm run build` to compile web components.
```
melaniebmn, line 28: Should the `s` at the end of `web packages` be removed in the sentence below?
Otherwise, people might think that there is more than one web package.
Can you also capitalize Angular and React, please?

packages/web/README.md
```diff
@@ -1,4 +1,4 @@
-[La version française suit.](#système-de-design---composants-gc)
+[La version française suit.](#composants-de-système-de-design-gc--web)
 
 # GC Design System Components – Web
```
melaniebmn, line 3: Flagging for discussion: should we remove the `- Web` part from the heading since the web package is GC Design System Components and not just another package?

melaniebmn (changes requested):
Added a few small comments.
Do we need to add any new instructions for icons to the React or Angular instructions?

packages/web/README.md
ethanWallace: Yeah looking at it more, I think this is a good idea.

ethanWallace:
> Do we need to add any new instructions for icons to the React or Angular instructions?

I am wondering if we should add a "Additional dependencies" section under installation that mentions what fonts are used and the Font awesome package we are using. We can even have a code block with the imports to help quickly add them to a project but people can add them anyway they want.

melaniebmn:
> > Do we need to add any new instructions for icons to the React or Angular instructions?
> 
> I am wondering if we should add a "Additional dependencies" section under Installation that mentions what fonts are used and the Font awesome package we are using. We can even have a code block with the imports to help quickly add them to a project but people can add them anyway they want.

I like that idea, and then we can do it the same way in the installation instructions.

melaniebmn (approved):
LGTM

PR #119, opened 2023-03-15
Update react package

ethanWallace:
# Summary | Résumé

Followed new instructions for React integration on Stenciljs documentation site to see if it would help with easier integration in other React frameworks. Didn't seem to what we wanted but has updated the React dependencies and cleaned up the package.


melaniebmn (approved):
That's too bad but still great to have this updated! LGTM!

PR #124, opened 2023-03-22
add gcds-container component + integrate it into other components

melaniebmn:
# Summary | Résumé

Added a `gcds-container` component and integrated it into these existing components:
- gcds-alert
- gcds-grid
- gcds-phase-banner


ethanWallace (approved):
LGTM

PR #125, opened 2023-03-22
Change gcds-components build to use gcds-tokens package

ethanWallace:
# Summary | Résumé

Switch away from using token files pushed by GitHub actions from `gcds-tokens` repo to using `@cdssnc/gcds-tokens` package.

`@cdssnc/gcds-components` now uses variables file from `@cdssnc/gcds-tokens` as global CSS instead of `global.css` file compiled using gulp. Gulp and the gulpfile have been stripped out of the project.


melaniebmn (approved):
LGTM

PR #130, opened 2023-03-23
[docs] fixes case

mnigh:
# Summary | Résumé

Fixes case of path in French section of `README.md` to match English equivalent.


melaniebmn (approved):
Thanks Matt for this suggestion, I will merge the change in :)

PR #135, opened 2023-03-28
Modify form components aria-describedby for better a11y

ethanWallace:
# Summary | Résumé

When testing form components, some configurations would cause the components `aria-describedby` attribute to render as `aria-describedby=" error-message-id"`. This was causing VoiceOver to not read the content linked by `aria-describedby`.

To fix this, I have changed how the `aria-describedby` renders. That should allow the content and the form input to be linked properly.

Also fixed a bug where the file summary in `gcds-file-uploader` would not be associated with the file input unless there was a `hint` or `error-message` on the component.


melaniebmn (approved):
LGTM

PR #144, opened 2023-04-17
Add error summary component

ethanWallace:
# Summary | Résumé

Add `gcds-error-summary` component to component library.

## Use

```html
<gcds-error-summary
  listen="true"        // boolean
  error-links=""      // string/object
  heading=""          // string
  sub-heading=""  // string
>
</gcds-error-summary>
```

### Listen

Using the `listen` attribute will listen for the `gcdsError` event and generate/display the error summary list when the parent form submits.

### Error-links

Use the `error-links` to render the error list without using event listeners. Pass a string or object formatted like the example below to render the list.
```
{
  "href": "Error message",
  "href": "Error message"
}
```

### Heading

To customize the error summary heading, pass a string to the `heading` attribute.

### Sub-heading

To customize the sub-heading, pass a string to the `sub-heading` attribute.


packages/web/src/components/gcds-error-summary/gcds-error-summary.tsx
```diff
@@ -0,0 +1,200 @@
...
+      }, 100);
+    }
+  }
+
+  /*
+  * Sort error object based on the order form compoennts appear in the form
+  */
+  sortErrors() {
+    let sortable = [];
+      for (var id in this.errorLinksObject) {
```
melaniebmn, line 118: Can we avoid using `var` and use `let` or `const` instead?

packages/web/src/components/gcds-error-summary/gcds-error-summary.tsx
```diff
@@ -0,0 +1,200 @@
...
+          tabindex="-1"
+          ref={element => this.shadowElement = element as HTMLElement}
+          class={`gcds-error-summary ${Object.keys(errorQueue).length > 0 ? 'gcds-show' : ''}`}
+        >
+          <h2 class="summary__heading">
+            {heading ?
+              heading
+            :
+              i18n[lang].heading
+            }
```
melaniebmn, line 176: Could we replace this with `{heading ?? i18n[lang].heading}`?

packages/web/src/components/gcds-error-summary/gcds-error-summary.tsx
```diff
@@ -0,0 +1,200 @@
...
+          </h2>
+          <p class="summary__sub-heading">
+            {subHeading ?
+              subHeading
+            :
+              i18n[lang].subheading
+            }
```
melaniebmn, line 183: Could we replace this with `{subHeading ?? i18n[lang].subheading}`?

melaniebmn (changes requested):
Can the error summary links scroll to the label for each field instead of the input field itself? I think that would be a slightly nicer interaction.

Aside from that, a couple of small comments but overall it looks great, nice work!

ethanWallace:
> Can the error summary links scroll to the label for each field instead of the input field itself? I think that would be a slightly nicer interaction.

Good idea! I will look at adding this!

melaniebmn (approved):
LGTM!

PR #158, opened 2023-05-01
Fix error summary triggering before submit

ethanWallace:
# Summary | Résumé

Add `hasSubmitted` state to error summary to fix triggering before `submit` event.

Also some React build files.


melaniebmn (approved):
LGTM

PR #159, opened 2023-05-02
remove boxshadow from focus state to support a line break

nmakuch:
### Removing box-shadow on error links 
- Noticing that some error messages will probably be long enough in some cases to force a line break and looks bad with the box-shadow. 
- The box-shadow needs to be removed from the focus style to account for this (and the error summary should be used on a white background anyway so removing it won't really matter)

| Before | After |
|--------|-------|
|   Screen Shot 2023-05-01 at 20 22 51 |  Screen Shot 2023-05-01 at 20 23 10 |


ethanWallace (approved):
LGTM!

PR #165, opened 2023-05-09
Update compile-and-publish.yml to allow triggering manually

ethanWallace:
# Summary | Résumé

Update the publish packages action to allow the action to be triggered manually.


melaniebmn (approved):
LGTM

PR #168, opened 2023-05-15
feat: publish new packages to CDN

patheard:
# Summary
Update the publish package workflow so that new packages are now also published to the CDN.

Each new package will be published to a versioned directory in the CDN and replace the files for its `@latest` version.  This will allow users to either link to versioned assets or the latest files.  

The flow for the publish step is as follows:

1. Use the `steps.publish.outputs.id` output to determine if a package was published.  The [JS-DevTools/npm-publish action](https://github.com/JS-DevTools/npm-publish#output) only provides this if there was a release in the format of `${name}@${version}`.
2. Sets up AWS credentials to publish to the CDN's S3 bucket.
3. Creates a `./tmp` directory and uses `npm install "$PUBLISHED_PACKAGE"` to download the package's assets.  This is done to make sure the npm package always matches the CDN assets.
4. Publishes the package assets to a `${name}@${version}` directory and `${name}@latest` directory on the CDN.

Part of this work for this change involved refactoring the workflow to use a matrix strategy so that the CDN publish logic did not have to be repeated for each package.

# Related
- cds-snc/platform-core-services#322

### DIFF
--- a/.github/workflows/compile-and-publish.yml
+++ b/.github/workflows/compile-and-publish.yml
@@ -69,22 +69,19 @@ jobs:
           role-session-name: CDNPublish
           aws-region: ${{ env.CDN_REGION }}
 
-      - name: Update CDN with published package
+      - name: Update CDN ${{ matrix.name }}
         if: steps.publish.outputs.id != ''
         run: |
           PUBLISHED_PACKAGE="${{ steps.publish.outputs.id }}"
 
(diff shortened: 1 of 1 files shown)


melaniebmn (approved):
LGTM - thank you for setting this up @patheard!

melaniebmn (approved):
LGTM

PR #170, opened 2023-05-17
feat: Slack notify on publish workflow failure

patheard:
# Summary
Add a workflow step that posts a message to the Design System ops channel when the publish workflow fails.

# Related
- cds-snc/platform-core-services#333


melaniebmn (approved):
LGTM

PR #172, opened 2023-05-18
Include i18n for all components

melaniebmn:
# Summary | Résumé

- Added localization files to all components with localization
- Removed padlock text from gcds-verify-banner

Note: While I was adding the localization file for the verify banner, I decided to remove the text referring to the padlock since browsers are removing the padlock soon.


ethanWallace (changes requested):
It's looking good, thank you for doing this!

Just one thing needs to be done before this can be merged. Would you be able to remove the `lock.svg` from the `gcds-verify-banner/assets` folder and it's import on `line 7` in the verify banner?

After that it's all good!

melaniebmn:
@ethanWallace good catch, thank you! PR is updated and the lock icon removed from the verify-banner.

ethanWallace (approved):
LGTM!

PR #175, opened 2023-05-24
update unpkg links to use new cdn links in CDN instructions

melaniebmn:
# Summary | Résumé

Replace unpkg links with new CDN links for CDN installation instructions.


ethanWallace (approved):
LGTM!

PR #179, opened 2023-05-30
Renovate base branch update

ethanWallace:
# Summary | Résumé

With the new `staging` branch, configure renovate to use the staging branch instead of main.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -7,5 +7,6 @@
     "schedule": [
       "before 6am on the first day of the month"
     ]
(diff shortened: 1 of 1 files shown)


patheard (approved):
Just double checking that you now only want Renovate to run against the `staging` branch and stop raising dependency PRs against `main`.  

If so, this'll do the trick!

ethanWallace:
Thanks @patheard, that's exactly what we want to do!

PR #180, opened 2023-05-31
gcds-lang-toggle + gcds-pagination stories

ethanWallace:
# Summary | Résumé

Add `gcds-lang-toggle` and `gcds-pagination` to storybook.

### Simple pagination rendering

`gcds-pagination` will no longer render the previous or next button if no `previous-href` or `next-href` is passed.

### URL property

The `url` property has been modified to accept and `object` or `string` like the `subLinks` property on the `gcds-footer`.


melaniebmn (approved):
LGTM

PR #182, opened 2023-06-14
Update publish github action to use jobs

ethanWallace:
# Summary | Résumé

To fix build/publish errors when running the `compile-and-publish.yml` action, switch the action to run sequential jobs.

The first job, `publish-web` will publish the `@cdssnc/gcds-components` npm package and publish the the CDN files. The second job, `publish-react-angular`, will do the same with the React and Angular packages but only if the `publish-web` job completes.

This change was required since the React and Angular packages are dependant on the new published version of `@cdssnc/gcds-components` and would fail if the web package was not published first.

### DIFF
--- a/.github/workflows/compile-and-publish.yml
+++ b/.github/workflows/compile-and-publish.yml
@@ -16,8 +16,8 @@ env:
   CDN_REGION: ca-central-1
 
 jobs:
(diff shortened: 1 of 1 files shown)


melaniebmn (approved):
LGTM

PR #183, opened 2023-06-22
Gcds-card component

ethanWallace:
# Summary | Résumé

Add new `gcds-card` to component library.

## Gcds-card

```
<gcds-card
  type="link"                // link | action
  card-title=""              // *required, string
  href=""                       // *required, string
  title-element="a"     // h3 | h4 | h5 | h6 | a


  description=""          // string
  tag=""                       // string


  img-src=""                 // string
  img-alt=""                 // string
>
 <slot name=”footer”></slot>
</gcds-card>
```

#### Link card

The link card is used to display information and has no call to action. The card footer is optional to display metadata about the card contents. If the card links to additional information the entire card is interactive.

#### Action card

The action card is used to display information and it contains a call to action in the card footer. Unlike the default card, only the headline link and the call to action in the card footer are interactive.

### DIFF
--- a/packages/web/package-lock.json
+++ b/packages/web/package-lock.json
@@ -18,7 +18,7 @@
       "devDependencies": {
         "@axe-core/puppeteer": "^4.3.2",
         "@babel/core": "^7.20.12",
-        "@cdssnc/gcds-tokens": "^1.1.1",
+        "@cdssnc/gcds-tokens": "^1.3.0",
         "@fortawesome/fontawesome-free": "^6.3.0",
         "@node-minify/core": "^6.2.0",
         "@node-minify/uglify-es": "^6.2.0",
@@ -1923,9 +1923,9 @@
       "dev": true
     },
     "node_modules/@cdssnc/gcds-tokens": {
-      "version": "1.1.1",
-      "resolved": "https://registry.npmjs.org/@cdssnc/gcds-tokens/-/gcds-tokens-1.1.1.tgz",
-      "integrity": "sha512-hkryLHgV/s50uOdc846568OZ9yC45DqQRja5NGuIrfdnE4lnL7nZbhSjvhGYLRtYx3e6oicYYoXxCMjLTsTZGw==",
+      "version": "1.3.0",
+      "resolved": "https://registry.npmjs.org/@cdssnc/gcds-tokens/-/gcds-tokens-1.3.0.tgz",
+      "integrity": "sha512-QmQLJpU8ePgEHI51TxvBpDAIJxIz7uB5IMmw3pAdJQzSwlwBtwFu0JzKZ5PRT2NW7AvumpZ5l3MUiggtu4O/rQ==",
       "dev": true
     },
     "node_modules/@cnakazawa/watch": {
@@ -28130,9 +28130,9 @@
       "dev": true
     },
     "@cdssnc/gcds-tokens": {
-      "version": "1.1.1",
-      "resolved": "https://registry.npmjs.org/@cdssnc/gcds-tokens/-/gcds-tokens-1.1.1.tgz",
-      "integrity": "sha512-hkryLHgV/s50uOdc846568OZ9yC45DqQRja5NGuIrfdnE4lnL7nZbhSjvhGYLRtYx3e6oicYYoXxCMjLTsTZGw==",
+      "version": "1.3.0",
+      "resolved": "https://registry.npmjs.org/@cdssnc/gcds-tokens/-/gcds-tokens-1.3.0.tgz",
+      "integrity": "sha512-QmQLJpU8ePgEHI51TxvBpDAIJxIz7uB5IMmw3pAdJQzSwlwBtwFu0JzKZ5PRT2NW7AvumpZ5l3MUiggtu4O/rQ==",
       "dev": true
     },
     "@cnakazawa/watch": {
--- a/packages/web/package-lock.json
+++ b/packages/web/package-lock.json
@@ -18,7 +18,7 @@
       "devDependencies": {
         "@axe-core/puppeteer": "^4.3.2",
         "@babel/core": "^7.20.12",
-        "@cdssnc/gcds-tokens": "^1.3.0",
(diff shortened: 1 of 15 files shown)


packages/web/package-lock.json
```diff
@@ -1923,9 +1923,15 @@
       "dev": true
     },
     "node_modules/@cdssnc/gcds-tokens": {
+<<<<<<< HEAD
+      "version": "1.3.0",
+      "resolved": "https://registry.npmjs.org/@cdssnc/gcds-tokens/-/gcds-tokens-1.3.0.tgz",
+      "integrity": "sha512-QmQLJpU8ePgEHI51TxvBpDAIJxIz7uB5IMmw3pAdJQzSwlwBtwFu0JzKZ5PRT2NW7AvumpZ5l3MUiggtu4O/rQ==",
+=======
```
melaniebmn, line 1930: Looks like there are some leftover merge conflicts in this file

packages/web/src/components/gcds-card/gcds-card.css
```diff
@@ -0,0 +1,128 @@
...
+        color: var(--gcds-card-title-color);
+        font: var(--gcds-card-title-font);
+        text-underline-offset: var(--gcds-card-title-text-underline-offset);
+        text-decoration-color: currentColor;
+        text-decoration-thickness: var(--gcds-card-title-text-decoration-thickness);
+        width: fit-content;
+      }
+  
+      .gcds-card__description {
+        max-width: 65ch;
```
melaniebmn, line 50: Can we use a container token here? The new Lg container token will be `62rem`, you can see a list of the changed tokens on this branch:

https://github.com/cds-snc/gcds-tokens/blob/add-nav-tokens/tokens/global/container/base.json

packages/web/src/components/gcds-card/gcds-card.tsx
```diff
@@ -0,0 +1,108 @@
...
+@Component({
+  tag: 'gcds-card',
+  styleUrl: 'gcds-card.css',
+  shadow: true,
+})
+export class GcdsCard {
+  @Element() el: HTMLElement;
+
+  /**
+   * The card title attribute specifies the title that appears on the card
```
melaniebmn, line 12: Seems like this description needs to be updated for the type attribute.

packages/web/src/components/gcds-card/gcds-card.tsx
```diff
@@ -0,0 +1,108 @@
...
+   */
+  @Prop({ reflect: true }) cardTitle!: string;
+
+  /**
+   * The card title attribute specifies the title that appears on the card
+   */
+  @Prop() titleElement: 'h3' | 'h4' | 'h5' | 'h6' | 'a' = 'a';
```
melaniebmn, line 24: Seems like this description needs to be updated for the titleElement attribute.

melaniebmn, line 24: How do you feel about changing this prop to `titleTag` to be more consistent with the other `tag` prop?

packages/web/src/components/gcds-card/gcds-card.tsx
```diff
@@ -0,0 +1,108 @@
...
+   */
+  @Prop({ reflect: true }) tag: string;
+
+  /**
+   * The img src attribute specifies the path to the image
+   */
+  @Prop({ reflect: true }) imgSrc: string;
+
+  /**
+   * The img alt attribute specifies the alt text for the image provided, if none, image will be decorative
```
melaniebmn, line 47: Should we add "and will not be read out to screen reader users" or something like that?

melaniebmn (changes requested):
Added a few comments, but looking good otherwise!

packages/web/src/components/gcds-card/gcds-card.tsx
ethanWallace: When first creating the property I was going to go with `title-tag` but changed it so it won't be confused with the `tag` property already available with the `gcds-card`.

That being said, I am willing to change it to `title-tag` if we think that is a better name for the property.

packages/web/src/components/gcds-card/gcds-card.tsx
ethanWallace: I don't think that will be necessary, saying the image will be decorative should cover that assistive technology won't pick it up.

packages/web/src/components/gcds-card/gcds-card.css
ethanWallace: I set the description width to `65ch` to match how we limit the line length for accessibility since the `description` could be a couple lines long.

Do you think it is still worth it to switch to a container token?

melaniebmn: Oh sorry, my bad! I read `65rem` instead of `65ch` - definitely keep it as `65ch` pls, but can we still put it in a token?

packages/web/src/components/gcds-card/gcds-card.tsx
melaniebmn: Mmh maybe that is an Amy question haha Let's keep as is and maybe check with Amy next week when she is back.

packages/web/src/components/gcds-card/gcds-card.tsx
```diff
@@ -0,0 +1,108 @@
...
+@Component({
+  tag: 'gcds-card',
+  styleUrl: 'gcds-card.css',
+  shadow: true,
+})
+export class GcdsCard {
+  @Element() el: HTMLElement;
+
+  /**
+   * The type attribute specifies the how the card renders as a link
```
melaniebmn, line 12: One small typo - there seems to be an extra `the` in `the how the`.

melaniebmn (changes requested):
2 small things and then it should be good to go!

packages/web/src/components/gcds-card/gcds-card.css
ethanWallace: Will do!

melaniebmn (approved):
LGTM

PR #194, opened 2023-07-14
Update storybook version to fix GitHub actions

ethanWallace:
# Summary | Résumé

To solve GitHub actions falling to install `@cdssnc/gcds-components` update storybook to use `v7.1.0-rc2`. 

Versions `7.0.*` were all causing a installation failure with `esbuild`.


melaniebmn (approved):
LGTM

Issue #204, opened 2023-07-21
gcds-checkbox change event not emitting correct value


j-fares:
The onChange function in the [GcdsCheckbox](https://github.com/cds-snc/gcds-components/blob/edcb6550f42f14275043cea96b6d4ddd9b7dd35c/packages/web/src/components/gcds-checkbox/gcds-checkbox.tsx#L278) class is emitting the value property (used for input) instead of the checked property.  This is causing the value of CustomEvent.detail to be undefined, instead of containing the most recent value of the checkbox, making the gcdsChange event unusable.

**Current Code Issue**
```typescript
 private onChange = () => {
    this.checked = !this.checked;

    this.gcdsChange.emit(this.value);
  };
```

**With Fix**

```typescript
 private onChange = () => {
    this.checked = !this.checked;

    this.gcdsChange.emit(this.checked);
  };
```

ethanWallace:
Thanks for flagging this. We have included this change in our `v0.12.0` release.

Issue #174, opened 2023-05-19
Vite JS Production Build


eliyammine:
Hi,

I have been developing in dev mode in Vite with no issues with the library. When attempting to build-deploy I am getting these errors:

![image](https://github.com/cds-snc/gcds-components/assets/6644807/39ed8131-ea43-4f69-97fb-a656cdbc335c)

I have done some research and these are some relevant issues I found:
https://github.com/ionic-team/stencil/issues/2827
https://github.com/vitejs/vite-plugin-react/issues/89

There appears to a build-flag in Stencil that should make this error go away:
https://stenciljs.com/docs/config-extras#enableimportinjection

Eli

eliyammine:
Spoke too soon on this one. I found a workaround via the config again.

1. npm i vite-plugin-dynamic-import -D
2. In vite.config.ts add the following:
```
import dynamicImport from 'vite-plugin-dynamic-import';

plugins: [
	dynamicImport({
		filter(id) {
			if (id.includes('/node_modules/@cdssnc')) {
				return true;
			}
		},
	}),
],
```

Will just leave this here incase anyone else experiences it

verena-ifx:
is there any update on this?

PR #205, opened 2023-07-24
Extra config + checkbox fix

ethanWallace:
# Summary | Résumé

Small adjustments for better use.

- Add `extra` config option `experimentalImportInjection` to stencil to help using the GCDS components in bundlers such as Vite. Will update to `enableImportInjection` when using `v3` of Stencil. #174 

- As flagged in #204, switch `onChange` on the `gcds-checkbox` to emit the `checked` value instead of `value`.


melaniebmn (approved):
LGTM

PR #215, opened 2023-08-04
gcds-search border-radius style fix

ethanWallace:
# Summary | Résumé

Use new tokens to fix the `border-radius` of the `gcds-search` component. The `border-end-end-radius` was overwritten by the `border-radius`.


melaniebmn (approved):
LGTM

PR #221, opened 2023-08-15
Tokens review update

nmakuch:
## This PR contains the following updates

These are changes to align with the figma file more and to make state styles more consistent

### File uploader updates
- Upload button had only default state when it should function like a secondary button
- Focus styles have been updated on the uploaded file and the remove button
- The remove button hover style was using `border-block-end` which was causing the uploaded file bucket to resize on state changes, it has been updated to use `text-decoration` instead.

| Before |  file-upload-before |
|--------|---|
| After  |  file-upload-after |

### Pagination updates
- hover text and border color has been applied to the hover state for pages
- `text-decoration` hidden on focus and active state to match some of our other states
- `pointer-events` removed from current page 
- Subheading on simple pagination was receiving different styling than the top line when they should be identical

| Before | pagination-before|
|--------|---|
| After  | pagination-after |

### DIFF
--- a/packages/web/package-lock.json
+++ b/packages/web/package-lock.json
@@ -21,7 +21,7 @@
         "@babel/core": "^7.20.12",
         "@babel/preset-env": "^7.20.2",
         "@babel/preset-typescript": "^7.21.0",
-        "@cdssnc/gcds-tokens": "^1.5.2",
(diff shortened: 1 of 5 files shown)


melaniebmn (approved):
LGTM

PR #224, opened 2023-08-22
signature storybook events and properties page

nmakuch:
### Creating storybook events and properties page for the signature component
- I'm working on the signature component pages for the doc site and the code tab needs to use the storybook component builder thingy
- Created a simple story in which you can toggle the language and type of the signature component

signature storybook


ethanWallace (changes requested):
Looking good so far. Can we also create the stories for the playground page and the overview page and the overview page itself?

PR #235, opened 2023-08-29
Fix focusout logic in gcds-topic-menu

ethanWallace:
# Summary | Résumé

Add additional logic checking to see if the theme and topic menu is open in the `focusout` listener. This prevents the menu from being toggled open if the closed menu button had focus on focus out.


melaniebmn (approved):
LGTM

PR #271, opened 2023-09-06
Remove minify script + dependencies

ethanWallace:
# Summary | Résumé

Remove minify script since it didn't make any difference with the built in minify from Stencil.


melaniebmn (approved):
LGTM

PR #272, opened 2023-09-06
Update media query in gcds-pagination

ethanWallace:
# Summary | Résumé

Update first media query in `gcds-pagination` to happen earlier to prevent overflow with page container.


melaniebmn (approved):
LGTM

PR #277, opened 2023-09-06
Update jest

ethanWallace:
# Summary | Résumé

For compatibility, update jest packages to `v27.0.3`. Won't be able to update to the latest version of jest until stencil is compatible with it.


melaniebmn (approved):
LGTM

PR #281, opened 2023-09-07
Add Eslint to components

melaniebmn:
# Summary | Résumé

Adding Eslint to components to improve code quality and add additional accessibility tests.

### DIFF
--- a/packages/web/scripts/copyCSSFile.js
+++ b/packages/web/scripts/copyCSSFile.js
@@ -1,3 +1,4 @@
-import { copyFileSync } from 'fs';
+// eslint-disable-next-line no-undef
+const fs = require('fs');
 
-copyFileSync('./dist/gcds/gcds.css', '../react/gcds.css');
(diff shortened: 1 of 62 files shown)


packages/web/scripts/copyCSSFile.js
```diff
@@ -1,3 +1,3 @@
-const fs = require('fs');
+import { copyFileSync } from 'fs';
```
ethanWallace, line 1: I receive a ` Cannot use import statement outside a module` error when running `npm run build` with these changes.

ethanWallace (changes requested):
It's looking good, thank you for adding this! Just one small thing I noticed.

packages/web/scripts/copyCSSFile.js
melaniebmn: Ah damn, that's too bad - I'll revert that change for now.

ethanWallace (approved):
LGTM!

PR #283, opened 2023-09-07
Add overview + playground pages for header and footer stories

ethanWallace:
# Summary | Résumé

Add the overview and playground storybook pages for the `gcds-header` and `gcds-footer`.


packages/web/src/components/gcds-footer/stories/gcds-footer.stories.tsx
```diff
@@ -65,11 +65,61 @@ const Template = (args) => (`
...
   display: 'full',
   contextualHeading: 'Contextual navigation',
   contextualLinks: '{ "Why GC Notify": "#", "Features": "#", "Activity on GC Notify": "#" }',
   subLinks: '',
   lang: 'en'
+};
+
+export const DisplayCompact = Template.bind({});
+DisplayCompact.args = {
+  lang: 'en'
```
melaniebmn, line 101: Can you add `display: 'compact'` here? It's currently showing up as `display="undefined"` in the code preview. Or if `compact` is supposed to be the default value, can you add it as the default value? Right now we don't have a default defined for `display`.

packages/web/src/components/gcds-header/stories/overview.mdx
```diff
@@ -0,0 +1,76 @@
...
+<br/>
+
+### Default
+
+<Canvas
+  of={Header.Default}
+  story={{ inline: true }}
+/>
+
+### Slots
```
melaniebmn, line 27: Could we maybe have an extra story here that shows all slots at once? Just so it's easier for the user to understand the order. For example, the menu slot is before the breadcrumbs slot, etc.

melaniebmn (changes requested):
Two small comments, otherwise looks good!

packages/web/src/components/gcds-footer/gcds-footer.tsx
```diff
@@ -17,14 +17,8 @@ export class GcdsFooter {
...
    */
-  @Prop({ reflect: true, mutable: true }) display!: 'compact' | 'full';
-
-  @Watch('display')
-  validateDisplay(newValue: string) {
-    if (newValue != 'compact' && newValue != 'full') {
-      this.display = 'compact';
-    }
-  }
+  @Prop({ reflect: true, mutable: true }) display: 'compact' | 'full' =
```
melaniebmn, line 20: Should we mark this prop as optional?

melaniebmn (approved):
Added one more question, but feel free to merge if you disagree :)

packages/web/src/components/gcds-footer/gcds-footer.tsx
ethanWallace: Good point, I will do that

melaniebmn (approved):
LGTM

PR #289, opened 2023-09-11
add story for button with icon

melaniebmn:
# Summary | Résumé

Very small PR... Adding a story to show an example of a button with an icon inside it.


ethanWallace (approved):
LGTM

PR #290, opened 2023-09-12
Add overview + playground pages for navigation components

ethanWallace:
# Summary | Résumé

Add the overview and playground pages to the navigation components in storybook.

### DIFF
--- a/packages/web/src/components/gcds-nav-group/stories/gcds-nav-group.stories.tsx
+++ b/packages/web/src/components/gcds-nav-group/stories/gcds-nav-group.stories.tsx
@@ -76,8 +76,102 @@ const Template = args =>
 </GcdsNavGroup>
 `.replace(/\s\snull\n/g, '');
 
+const TemplateTopNav = args =>
+  `
+<!-- Web component code (Angular, Vue) -->
+<gcds-top-nav label="Top nav example" alignment="right">
+  <gcds-nav-group
+    menu-label="${args.menuLabel}"
+    open-trigger="${args.openTrigger}"
+    ${args.closeTrigger ? `close-trigger="${args.closeTrigger}"` : null}
+    ${args.open ? `open` : null}
+    ${args.lang != 'en' ? `lang="${args.lang}"` : null}
+  >
+    <gcds-nav-link href="#">Nav link</gcds-nav-link>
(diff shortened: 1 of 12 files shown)


packages/web/src/components/gcds-nav-group/stories/gcds-nav-group.stories.tsx
```diff
@@ -84,3 +133,48 @@ Default.args = {
   open: false,
   lang: 'en',
 };
+
+export const Open = Template.bind({});
+Open.args = {
+  menuLabel: '',
```
melaniebmn, line 139: Same comment - can you remove `menuLabel` from the code preview if it's empty?

packages/web/src/components/gcds-nav-group/stories/gcds-nav-group.stories.tsx
```diff
@@ -84,3 +133,48 @@ Default.args = {
...
+  openTrigger: 'Navigation group',
+  closeTrigger: '',
+  open: true,
+  lang: 'en',
+};
+
+export const Triggers = Template.bind({});
+Triggers.args = {
+  menuLabel: '',
```
melaniebmn, line 148: Same comment - can you remove `menuLabel` from the code preview if it's empty?

packages/web/src/components/gcds-nav-group/stories/gcds-nav-group.stories.tsx
```diff
@@ -84,3 +133,48 @@ Default.args = {
...
+  menuLabel: 'Sub-menu',
+  openTrigger: 'Navigation group',
+  closeTrigger: '',
+  open: false,
+  lang: 'en',
+};
+
+export const TopNav = TemplateTopNav.bind({});
+TopNav.args = {
+  menuLabel: '',
```
melaniebmn, line 166: Same comment - can you remove `menuLabel` from the code preview if it's empty?

melaniebmn (changes requested):
Thank you for adding these - the last missing ones woho

I added a small comment and was also wondering if you could add a separate story for each props table instead of using the default story? Simply so that the default story will always stay the same, no matter what someone selects in the props table.

melaniebmn (approved):
Thanks for making those changes - LGTM

PR #291, opened 2023-09-12
Gcds-header: Remove label from nav landmark to avoid confusion

ethanWallace:
# Summary | Résumé

Remove the "skip to main content" label from the navigation landmark to avoid confusion when navigating the header using a screen reader.


melaniebmn (approved):
LGTM

PR #294, opened 2023-09-14
Small adjustments

melaniebmn:
# Summary | Résumé

- Signature: add default value for type and variant.
- Fix topic menu focus border for menu btn

@ethanWallace the `has-link` prop for the signature component currently doesn't have a default value, but the Storybook story shows the default value as `false`. Should we set the default value to `false` in the component as well?


ethanWallace:
Yes, let's set it to the default value in the component as well

melaniebmn:
@ethanWallace added it in!

ethanWallace (approved):
LGTM!

PR #305, opened 2023-10-10
feat: Link component

daine:
# Summary | Résumé
Adds a new Link component to the Design System

- fixes https://github.com/cds-snc/design-gc-conception/issues/506

## Related documentation
- [Implementation and requirements
](https://docs.google.com/document/d/1d717T3S6J3qhq58jXTQ5cBuPfJMp2WWftfaU_imu81A/edit)
- [Website documentation document](https://docs.google.com/document/d/1hVmy8mPts5kMvM1Vz-nShS-zUXrvsvIN3wAdxMwVdaY/edit)

## Testing
I run the following command to frequently test this specific component:
`npm run test -- -- gcds-link`

## Introduces a link component. Here's a sample of what it looks like
Screenshot 2023-10-19 at 10 26 20 AM


packages/web/src/components/gcds-link/i18n/i18n.js
```diff
@@ -0,0 +1,16 @@
...
+  },
+  fr: {
+    external: "S'ouvre dans un nouvel onglet.",
+    phone: 'Phone number [fr & copy edit needed]',
+    download: 'Download [fr & copy edit needed]',
+    email: 'Email [fr & copy edit needed]',
+  },
+};
+
+export default I18N;
```
daine, line 16: @SmartMouthWords Hi Amy, I could use some help with the language for this. A bit of context: these are the labels for the icons when a link is either a download, an email, or a phone number.
- A download link will download the file to the user's device if the URL is of the same origin. If it's not, it simply opens in the same tab.
- A `mailto:` or email link will attempt to open the user's email client on their device
- A `tel:` link will attempt to open the user's phone app
- Both `mailto:` and `tel:` link actions can change depending on the user's setting, and we have no way of forcing it upon them.

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,220 @@
...
+      target,
+      download,
+      type,
+    };
+
+    const isExternal = target === '_blank' || external;
+
+    return (
+      <Host>
+        <Tag
```
ethanWallace, line 174: Since this will only render as an `a` tag, you can just switch this to `a` instead of `Tag`

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,220 @@
...
+      target,
+      external,
+      download,
+      type,
+      inheritedAttributes,
+    } = this;
+
+    const Tag = 'a';
+    const attrs = {
+      display,
```
ethanWallace, line 162: `display` isn't a HTML attribute so it won't apply the desired display to the attribute.

You will need to create a class to apply the desired `display` property to the `:host` element. An example of this can be found in the new Text component

packages/web/src/components/gcds-link/gcds-link.css
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  &
+  :hover {
+    background-color: var(--gcds-button-text-only-default-background);
+    color: var(--gcds-link-hover);
+    text-decoration-thickness: var(--gcds-button-text-only-hover-decoration-thickness);
+  }
+
+  &
```
ethanWallace, line 22: For these selectors to work, you will need to format the CSS like `&:focus`.  Right now the component doesn't render the proper styles on hover and focus.

packages/web/src/components/gcds-link/gcds-link.css
```diff
@@ -0,0 +1,43 @@
+:host a {
+  cursor: pointer;
+  color: var(--gcds-link-default);
+
+  background-color: var(--gcds-button-text-only-default-background);
```
ethanWallace, line 5: You should be able to remove this `background-color` line.

packages/web/src/components/gcds-link/gcds-link.css
```diff
@@ -0,0 +1,43 @@
...
+  &
+  :hover {
+    background-color: var(--gcds-button-text-only-default-background);
+    color: var(--gcds-link-hover);
+    text-decoration-thickness: var(--gcds-button-text-only-hover-decoration-thickness);
+  }
+
+  &
+  :focus {
+    background-color: var(--gcds-button-shared-focus-background);
```
ethanWallace, line 24: Can you use the tokens you created in https://github.com/cds-snc/gcds-tokens/pull/173?

ethanWallace (changes requested):
Just a few things I noticed

packages/web/src/components/gcds-button/gcds-button.css
```diff
@@ -86,38 +86,6 @@
...
-      border-color: var(--gcds-button-shared-focus-background);
-      text-decoration: none;
-    }
-
-    &:active {
-      background-color: var(--gcds-button-shared-active-background);
-      border-color: var(--gcds-button-shared-active-background);
-    }
-  }
-
```
melaniebmn, line 120: I can't comment on the right line cuz Github (:D) but for lines 169 + 170:
Since we won't be supporting the `text-only` style for buttons anymore, can we update these 2 lines to remove the complexity that is not needed anymore?

packages/web/src/components/gcds-link/gcds-link.css
```diff
@@ -0,0 +1,43 @@
+:host a {
+  cursor: pointer;
+  color: var(--gcds-link-default);
+
+  background-color: var(--gcds-button-text-only-default-background);
+
+  &
+  :not([aria-disabled='true']) {
+    text-decoration: underline;
+    text-underline-offset: 0.2em;
```
melaniebmn, line 10: Can you add a token for this, please? We are trying to avoid any hardcoded values in this repo.

packages/web/src/components/gcds-link/gcds-link.css
```diff
@@ -0,0 +1,43 @@
...
+    text-decoration-color: currentColor;
+    text-decoration-thickness: var(--gcds-button-text-only-default-decoration-thickness);
```
melaniebmn, line 12: Can you replace this with a link token instead of using a button token for this?

packages/web/src/components/gcds-link/gcds-link.css
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  &
+  :hover {
+    background-color: var(--gcds-button-text-only-default-background);
+    color: var(--gcds-link-hover);
+    text-decoration-thickness: var(--gcds-button-text-only-hover-decoration-thickness);
```
melaniebmn, line 19: Same comment for lines 17 + 19:
Can you replace the button token with a link token instead?

packages/web/src/components/gcds-link/gcds-link.css
melaniebmn: Applies to lines 24-29

packages/web/src/components/gcds-link/gcds-link.css
```diff
@@ -0,0 +1,43 @@
...
+    outline: var(--gcds-button-shared-focus-outline-width) solid var(--gcds-button-shared-focus-background);
+    outline-offset: var(--gcds-button-border-width);
+    border-color: var(--gcds-button-shared-focus-background);
+    text-decoration: none;
+  }
+
+  &
+  :active {
+    background-color: var(--gcds-button-shared-active-background);
+    border-color: var(--gcds-button-shared-active-background);
```
melaniebmn, line 36: Same comment for lines 35 + 36:
Can you replace the button token with a link token instead?

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,220 @@
...
+  @Watch('size')
+  validateSize(newValue: string) {
+    const values = ['regular', 'small'];
+
+    if (!values.includes(newValue)) {
+      this.size = 'regular';
+    }
+  }
+
+  // TODO: Do we need to validate display, rel, target, etc?
```
melaniebmn, line 44: Can this comment be removed?

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+  /**
+   * The rel attribute specifies the relationship between the current document and the linked document
+   */
+  @Prop() rel: string | undefined;
+
+  /**
+   * The target attribute specifies where to open the linked document
+   */
+  @Prop() target: string | undefined;
```
melaniebmn, line 74: Can we limit this to the available options that are available for the attribute:
`"_blank" | "_self" | "_parent" | "_top"`

And set the default value to `"_self"`?

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,220 @@
...
+    const values = ['regular', 'small'];
+
+    if (!values.includes(newValue)) {
+      this.size = 'regular';
+    }
+  }
+
+  // TODO: Do we need to validate display, rel, target, etc?
+
+  private handleClick = (e: Event) => {
```
melaniebmn, line 46: Can this function be moved out of the props section and at the bottom of the `Events` section?

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+    return (
+      <Host>
+        <Tag
+          {...attrs}
+          onClick={e => this.handleClick(e)}
+          class={`link--${size}`}
+          ref={element => (this.shadowElement = element as HTMLElement)}
+          target={isExternal ? '_blank' : target}
+          rel={isExternal ? 'noopener' : rel}
```
melaniebmn, line 180: Can we also add in `noreferrer` for external links to reduce the risk of reverse tab napping for external links?

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,220 @@
...
+          {...inheritedAttributes}
+          part="link"
+        >
+          <slot name="left"></slot>
```
melaniebmn, line 184: We can remove the slot, this was an icon slot for the button component that is not needed.

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,220 @@
...
+              margin-left="100"
+            />
+          ) : href.toLowerCase().startsWith('tel:') ? (
+            <gcds-icon
+              name="phone"
+              label={i18n[lang].phone}
+              margin-left="100"
+            />
+          ) : (
+            <slot name="right"></slot>
```
melaniebmn, line 214: We can remove the slot, this was an icon slot for the button component that is not needed.

packages/web/src/components/gcds-link/test/gcds-link.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+          <a class="link--regular" display="inline" href="https://google.com" part="link" download>
+            <slot name="left"></slot>
+            <slot></slot>
+            <slot name="right"></slot>
+          </a>
+        </mock:shadow-root>
+        Download file
+      </gcds-link>
+    `);
+  });
```
melaniebmn, line 64: Can you please also add variant tests in french so we can check that the labels get translated correctly?

melaniebmn: Maybe a test for the smaller size would be nice as well.

packages/web/src/index.html
```diff
@@ -374,6 +376,19 @@ <h4 class="mt-500 mb-300">Variable width icons (auto width)</h4>
...
+      <div class="my-200"><gcds-link>Base link</gcds-link></div>
+      <div class="my-200">This is a <gcds-link size="small" href="/">small link</gcds-link> if you compare it to regular text.</div>
+      <div class="my-200"><gcds-link href="https://picsum.photos/480/270" external>This is an external link</gcds-link></div>
+      <div class="my-200">This is a <gcds-link href="https://picsum.photos/200/200.jpg" download="random_image" type="image/jpeg">download link (PDF, 1.3 MB)</gcds-link> with type indicated</div>
+      <div class="my-200">Contact us <gcds-link href="mailto:test@test.com?subject=Test%20email">via email</gcds-link> or by phone.</div>
+      <div class="my-200">Contact us <gcds-link href="tel:1234567890">by phone</gcds-link> or by email.</div>
+      <div class="my-200">There is text before and <gcds-link href="/" display="inline-block">(sample link)</gcds-link> after this link.</div>
+      <div class="my-200"><gcds-link href="/" referrerpolicy="no-referrer" external>This is a sample link with a referrerpolicy</gcds-link></div>
+      <div class="my-200"><gcds-link href="" onClick='window.alert("Thank you!")'>Click me!</gcds-link></div>
```
melaniebmn, line 389: This is suuuuuper nit-picky so apologies about that but can you update all these div's to use p tags instead since these are all text-based links? Trying to use semantic tags whenever possible :)

packages/web/src/components/gcds-link/stories/overview.mdx
```diff
@@ -0,0 +1,63 @@
...
+<Canvas
+  of={Link.Phone}
+  story={{ inline: true }}
+/>
+
+### Size
+<Canvas
+  of={Link.Sizes}
+  story={{ inline: true }}
+/>
```
melaniebmn, line 48: Could you add another example here of a regular-sized link, please? Showing both options makes it a bit easier to see the difference.

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,220 @@
...
+  private shadowElement?: HTMLElement;
+
+  /**
+   * Props
+   */
+
+  /**
+   * Set the link size
+   */
+  @Prop({ mutable: true }) size: 'regular' | 'small' = 'regular';
```
melaniebmn, line 33: @daine @ethanWallace wondering if we should add `inherit` as a third option and set the `size` prop to default to `inherit`.

Reasoning for this:

The `regular` size currently gets its font size from the parent element, which works well for links inside a text block. However, when these links are used outside of a text block context (e.g. Resources links on the Storybook link overview page), they still inherit the font size from the parent. In this case, the links appear smaller than the expected regular font size.

To avoid this, we could add the `inherit` option as the default value for the `size` prop. The `inherit` option can then be used for links within headings or other text blocks. The `regular` option can be used for standalone links (e.g. the links in the resources section) and use a font token to load our regular text styling.

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,216 @@
...
+      control: { type: 'radio' },
+      options: ['regular', 'small'],
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: 'regular' },
+      },
+    },
+    display: {
+      control: 'select',
+      options: ['inline', 'inline-block', 'block'],
```
melaniebmn, line 23: Can you set `inline` as the default value, please?

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,216 @@
...
+        type: { summary: 'string' },
+        defaultValue: { summary: '-' },
+      },
+    },
+    target: {
+      control: 'select',
+      options: ['_blank', '_self', '_parent', '_top'],
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: '-' },
```
melaniebmn, line 44: Can you set the default value to `_self`, please?

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,216 @@
...
+      },
+    },
+    type: {
+      control: { type: 'text' },
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: 'button' },
```
melaniebmn, line 65: We don't seem to have a default value defined for the type, can you update the default value her to `-`?

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,216 @@
...
+} ${args.href ? `href="${args.href}"` : null
+} ${args.rel ? `rel="${args.rel}"` : null
+} ${args.target ? `target="${args.target}"` : null
+} ${args.size != 'regular' && args.size ? `size="${args.size}"` : null
+} ${  args.external ? `external` : null
+} ${args.download ? `download="${args.download}"` : null
+} ${args.type ? `type="${args.type}"` : null
+}>  ${args.default}
+</gcds-link>
+  ${args.default}
```
melaniebmn, line 131: Lines 131 + 132 seem to be duplicates and could be removed.

melaniebmn (changes requested):
Really nice work getting all the different pieces set up and figured out :) 

I added some comments, please let me know if you have questions about anything.

For the button component:
We will need to make a few more changes in order to remove the link styling from there. I'll add the list here. Feel free to make these changes or let me know and we can do them in a separate PR:
- The `buttonStyle` prop needs to be removed from the button code in `gcds-button.tsx`. It also needs to be removed from the button stories as well as the applicable button tests.
- The `left` and `right` slots which can be removed for the link, and can also be removed for the button as well.

packages/web/src/components/gcds-link/gcds-link.tsx
ethanWallace: Good idea! I am for it

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,220 @@
...
+    if (!values.includes(newValue)) {
+      this.size = 'regular';
+    }
+  }
+
+  // TODO: Do we need to validate display, rel, target, etc?
+
+  private handleClick = (e: Event) => {
+    if (this.clickHandler) {
+      this.clickHandler(e);
```
ethanWallace, line 48: Forgot this one last thing, can we remove the clickHandler? It's one thing I want to go through all the components and redo so no need to have it here. 

When a click event happens we will need to emit the custom events like `this.gcdsClick.emit();`.

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,220 @@
...
+      type,
+    };
+
+    const isExternal = target === '_blank' || external;
+
+    return (
+      <Host>
+        <Tag
+          {...attrs}
+          onClick={e => this.handleClick(e)}
```
ethanWallace, line 176: Can we also handle the `onFocus` and `onBlur` and emit the custom events from them?

ethanWallace (commented):
Just a few things that came to mind around events

packages/web/src/components/gcds-link/gcds-link.css
daine: Thanks, just noticed I think my editor is not configured properly for CSS; don't know why it reformatted my file to this!

packages/web/src/components/gcds-link/i18n/i18n.js
daine: > Where are all the places this content will live?

These will be the labels for screen readers (`aria-label`). 
Screenshot 2023-10-25 at 4 32 12 PM

> Do you want me to look at the parts labelled copy edit? Or at the whole page?

The copy edits would be great!

> "This is a download link [...] with type indicated", you might leave off "with type indicated". This isn't stated for the others, like phone so may not need to be here. If you think the download icon is less recognizable (probably is), you could add a line about it but then you'd probably want to do the same for the external link icon.


I think I need further clarity on what you mean, let's chat about this soon!

> Could the text "Referrerpolicy" be replaced with "policy reference"?

No, that is an HTML attribute set by W3 standards. Unless I'm missing some context?

> For the labels:
>
> For external, is it always a new tab, could it be a window or something else?
For the email one, you might use "default email" or "default email app" instead of "email client".
Recommend instead of using "may" communicating what will happen in what circumstances and potentially removing "your". For example, instead of "May open your phone app", you could consider stating:
"Will open the phone app that's connected" or
"Will open the phone app. If no phone app is connected, link will be disabled." or
"Will open the phone app or prompt to copy the number"
(Not sure how this will work so included a few examples based on guesses.)

Currently, when this was implemented as a button link, that's the text we have (for external links, it says opens in a new tab). It could open in a new tab or window (from [Mozilla a tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target): usually a new tab, but users can configure browsers to open a new window instead.)

packages/web/src/components/gcds-link/gcds-link.tsx
daine: Sounds good, doing that actually brought a few a11y issues so I've done a few updates!

packages/web/src/components/gcds-link/gcds-link.css
daine: Just saw this too late, but yes definitely!

packages/web/src/components/gcds-link/gcds-link.tsx
daine: Oh I'm actually asking this question for the PR haha. I'm guessing no to all?

packages/web/src/index.html
daine: Done! No worries about the nit

packages/web/src/components/gcds-link/gcds-link.tsx
daine: I think so! @melaniebmn I know we talked about keeping the click handler, would you mind if we took it out?

daine: @melaniebmn also apologies if I'm misunderstanding what we talked about and just mistakenly kept the clickhandler

melaniebmn: All good, you can take it out @daine :)

packages/web/src/components/gcds-link/gcds-link.tsx
melaniebmn: Oops sorry, didn't get that :D I would say, whenever you have predefined options for a prop, it's good to validate! We don't do this for all components yet, it's something that we started later but we want to add it for all components.

packages/web/src/components/gcds-link/gcds-link.tsx
daine: For future reference to myself, found the code here: https://github.com/cds-snc/gcds-components/blob/ac5d6e7db5b0ec3155441ff33d243c27d6322854/packages/web/src/components/gcds-text/gcds-text.tsx#L184

packages/web/src/components/gcds-link/gcds-link.tsx
daine: What about the user-defined value for "framename" when they're directing to a named iframe? @melaniebmn @ethanWallace

packages/web/src/components/gcds-link/gcds-link.tsx
daine: This is what I'm planning on doing to remove this extra slot, but wondering if we could make the code better.
```suggestion
          ) : (
            href.toLowerCase().startsWith('tel:') && (
              <gcds-icon
                name="phone"
                label={i18n[lang].phone}
                margin-left="100"
              />
            )
          )}
```

packages/web/src/components/gcds-link/gcds-link.tsx
daine:
```suggestion
  /**
   * The href attribute specifies the URL of the page the link goes to
   */
  @Prop() target: string = "_self";
```

packages/web/src/components/gcds-link/gcds-link.tsx
daine: If you could send me a sample that be be awesome @melaniebmn

packages/web/src/components/gcds-link/test/gcds-link.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+            <gcds-icon name="external-link" label="Opens in a new tab." margin-left="200" />
+          </a>
+        </mock:shadow-root>
+        External Link
+      </gcds-link>
+    `);
+  });
+
+  // TODO: add variants for testing french labels
+  // examples where?
```
daine, line 48: Examples here: https://github.com/cds-snc/gcds-components/blob/develop/packages/web/src/components/gcds-date-modified/test/gcds-date-modified.spec.tsx#L26

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
daine: Reminder for myself: set the events & properties tab to use the default values

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+  /**
+   * Emitted when the link has focus.
+   */
+  @Event() gcdsFocus!: EventEmitter<void>;
+
+  /**
+   * Emitted when the link loses focus.
+   */
+  @Event() gcdsBlur!: EventEmitter<void>;
```
daine, line 126: @ethanWallace can I remove these?

ethanWallace: It would be best if the `gcds-link` emits those events on focus and on blur and even emit a click event as well. 

You would need something like the code below on the `a` tag. Might not be 100% correct, might have to play around with it.

``` jsx
<a
  onBlur={this.gcdsBlur.emit()}
  onFocus={this.gcdsFocus.emit()}
  onClick={this.gcdsClick.emit()}
```

daine:
Added a new token here - https://github.com/cds-snc/gcds-tokens/pull/185

packages/web/src/components/gcds-link/gcds-link.tsx
daine: I'm looking around and found this, I'll try it out, thanks Ethan!
https://github.com/cds-snc/gcds-components/blob/develop/packages/web/src/components/gcds-search/gcds-search.tsx#L137-L139

daine:
> For the button component: We will need to make a few more changes in order to remove the link styling from there. I'll add the list here. Feel free to make these changes or let me know and we can do them in a separate PR:
> 
> * The `buttonStyle` prop needs to be removed from the button code in `gcds-button.tsx`. It also needs to be removed from the button stories as well as the applicable button tests.
> * The `left` and `right` slots which can be removed for the link, and can also be removed for the button as well.

I attempted to cleanup the button component text-only aspect, but I think it's a bigger task so it would be better to do it in a separate PR and issue

packages/web/src/components/gcds-link/gcds-link.css
```diff
@@ -0,0 +1,30 @@
...
+    text-decoration-thickness: var(--gcds-link-decoration-thickness);
+  }
+
+  &:hover {
+    color: var(--gcds-link-hover);
+    text-decoration-thickness: var(--gcds-link-hover-decoration-thickness);
+  }
+
+  &:focus {
+    background-color: var(--gcds-link-focus-background);
```
ethanWallace, line 18: I don't want to be annoying but would we be able to add a `border-radius` to the focus style to be consistent with the focus styles of other components.

melaniebmn:
> > For the button component: We will need to make a few more changes in order to remove the link styling from there. I'll add the list here. Feel free to make these changes or let me know and we can do them in a separate PR:
> > 
> > * The `buttonStyle` prop needs to be removed from the button code in `gcds-button.tsx`. It also needs to be removed from the button stories as well as the applicable button tests.
> > * The `left` and `right` slots which can be removed for the link, and can also be removed for the button as well.
> 
> I attempted to cleanup the button component text-only aspect, but I think it's a bigger task so it would be better to do it in a separate PR and issue

@daine sounds good, I can do that if you like!

packages/web/src/components/gcds-button/gcds-button.css
```diff
@@ -93,39 +93,6 @@
...
-      box-shadow: var(--gcds-button-shared-focus-box-shadow);
-      text-decoration: none;
-    }
-
-    &:active {
-      background-color: var(--gcds-button-shared-active-background);
-      border-color: var(--gcds-button-shared-active-background);
-    }
-  }
-
```
melaniebmn, line 128: Can you revert this change since we won't be doing the button changes in this PR anymore?

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,221 @@
...
+  @Watch('size')
+  validateSize(newValue: string) {
+    const values = ['regular', 'small', 'inherit'];
+
+    if (!values.includes(newValue)) {
+      this.size = 'regular';
+    }
+  }
+
+  @Watch('display')
```
melaniebmn, line 44: Can you move this below the `display` prop so that the prop + validation are grouped together?

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,221 @@
...
+  validateDisplay(newValue: string) {
+    const values = ['block', 'inline', 'inline-block'];
+
+    if (!values.includes(newValue)) {
+      this.display = 'block';
```
melaniebmn, line 49: Can you change this to `inline` please since we assigned `inline` as the default value for `block`?

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,221 @@
...
+    }
+  }
+
+  /**
+   * Link props
+   */
```
melaniebmn, line 55: Sorry super picky, but can we remove this comment? It was only needed in the button component because we had button and link-specific props so we can remove it here.

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+  /**
+   * The target attribute specifies where to open the linked document
+   */
+  @Prop() target?: string = '_self';
+
+  /**
+   * Whether the link is external or not
+   */
+  @Prop() external: false;
```
melaniebmn, line 80: Can you set the type to boolean and then the default value to false? And can we also flag it as optional prop?

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,221 @@
...
+  private shadowElement?: HTMLElement;
+
+  /**
+   * Props
+   */
+
+  /**
+   * Set the link size
+   */
+  @Prop({ mutable: true }) size: 'regular' | 'small' | 'inherit' = 'regular';
```
melaniebmn, line 33: Can you add a question mark after `size` to mark it as optional prop?

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  /**
+   * Link props
+   */
+
+  /**
+   * Sets the display behavior of the link
+   */
+  @Prop() display: 'block' | 'inline' | 'inline-block' = 'inline';
```
melaniebmn, line 60: Can you add a question mark after `display` to mark it as optional prop?

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+  /**
+   * The href attribute specifies the URL of the page the link goes to
+   */
+  @Prop() href!: string;
+
+  /**
+   * The rel attribute specifies the relationship between the current document and the linked document
+   */
+  @Prop() rel: string | undefined;
```
melaniebmn, line 70: Can you add a question mark after `rel` to mark it as optional prop?

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+  /**
+   * Whether the link is external or not
+   */
+  @Prop() external: false;
+
+  /**
+   * The download attribute specifies that the target (the file specified in the href attribute) will be downloaded when a user clicks on the hyperlink
+   */
+  @Prop() download: string | undefined;
```
melaniebmn, line 85: Can you add a question mark after `download` to mark it as optional prop?

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+  /**
+   * The type specifies the media type of the linked document
+   */
+  @Prop() type: string | undefined;
```
melaniebmn, line 90: Can you add a question mark after `type` to mark it as optional prop?

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,221 @@
...
+      if (mutations[0].oldValue != this.el.lang) {
+        this.lang = this.el.lang;
+      }
+    });
+    observer.observe(this.el, observerConfig);
+  }
+
+  componentWillLoad() {
+    // Validate attributes and set defaults
+    this.validateSize(this.size);
```
melaniebmn, line 135: Can you add `validateDisplay` here as well, please?

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,221 @@
...
+            />
+          ) : download !== undefined ? (
+            <gcds-icon
+              name="download"
+              label={i18n[lang].download}
+              margin-left="100"
+            />
+          ) : href.toLowerCase().startsWith('mailto:') ? (
+            <gcds-icon
+              iconStyle="regular"
```
melaniebmn, line 203: Can you switch this prop to use kebab case please to stay consistent aka `icon-style`?

packages/web/src/components/gcds-link/gcds-link.css
```diff
@@ -0,0 +1,30 @@
...
+    color: var(--gcds-link-focus-text);
+    outline: var(--gcds-link-focus-outline-width) solid var(--gcds-link-focus-background);
+    outline-offset: var(--gcds-link-focus-outline-offset);
+    border-color: var(--gcds-link-focus-border-color);
+    text-decoration: none;
+  }
+
+  /* Small link styling */
+  &.link--small {
```
melaniebmn, line 27: Please add support for `link--regular` and `link--inherit` as well.

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,221 @@
...
+    const isExternal = target === '_blank' || external;
+
+    return (
+      <Host>
+        {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions,jsx-a11y/click-events-have-key-events */}
+        <a
+          role="link"
+          tabIndex={0}
+          {...attrs}
+          class={`link--${size} ${display != 'block' ? `d-${display}` : ''}`}
```
melaniebmn, line 178: **Size**
The CSS currently only has a class for `link-small`, can we also add a class for `link-regular` and `link-inherit` to support those prop values as well?

**Block**
The CSS currently doesn't have any classes for the display prop. Can we add a class for `d-block`,  and `d-inline-block` to support those prop values? And then can we change the conditional to only add the class if it's not equal to `inline` since `inline` is supposed to be the default value?

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,211 @@
...
+};
+
+export const Download = Template.bind({});
+Download.args = {
+  href: 'test.jpg',
+  target: '_self',
+  size: 'regular',
+  download: 'file.pdf',
+  type: 'image/jpg',
+  default: 'a file download',
```
melaniebmn, line 181: I know this is just an example of a fake PDF download but to follow our own guidance can we add the file type + size in brackets to the link text, please?

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,211 @@
...
+      },
+    },
+  },
+};
+
+const Template = args =>
+  `
+This is an example of
+<!-- Web component code (Angular, Vue) -->
+<gcds-link ${args.display ? `display="${args.display}"` : null} ${
```
melaniebmn, line 88: In order to output only code in these examples that users would have to add to replicate the example, we try to only add the code if the value for a prop is not the default value. So for props that have default values, we add these props in the story template like this:
`${args.display != 'inline' ? `display="${args.display}"` : null}`

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,211 @@
...
+  } ${args.rel ? `rel="${args.rel}"` : null} ${
+    args.target ? `target="${args.target}"` : null
+  } ${args.size != 'regular' && args.size ? `size="${args.size}"` : null} ${
+    args.external ? `external` : null
+  } ${args.download ? `download="${args.download}"` : null} ${
+    args.type ? `type="${args.type}"` : null
+  }>${args.default}</gcds-link>
+
+<!-- React code -->
+<GcdsLink ${args.display ? `display="${args.display}"` : null} ${
```
melaniebmn, line 99: Same comment about the display prop, can we change that to the following to only have it displayed if it's not set to its default value?
```${args.display != 'inline' ? `display="${args.display}"` : null}```

File: packages/web/src/components/gcds-link/stories/gcds-link.stories.js [hunk 48]
```diff
@@ -0,0 +1,211 @@
...
+};
+
+const Template = args =>
+  `
+This is an example of
+<!-- Web component code (Angular, Vue) -->
+<gcds-link ${args.display ? `display="${args.display}"` : null} ${
+    args.href ? `href="${args.href}"` : null
+  } ${args.rel ? `rel="${args.rel}"` : null} ${
+    args.target ? `target="${args.target}"` : null
```
melaniebmn, line 91: Same comment about the target prop, can we change that to only have it displayed if it's not set to its default value?

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,211 @@
...
+    args.external ? `external` : null
+  } ${args.download ? `download="${args.download}"` : null} ${
+    args.type ? `type="${args.type}"` : null
+  }>${args.default}</gcds-link>
+
+<!-- React code -->
+<GcdsLink ${args.display ? `display="${args.display}"` : null} ${
+    args.href ? `href="${args.href}"` : null
+  } ${args.rel ? `rel="${args.rel}"` : null} ${
+    args.target ? `target="${args.target}"` : null
```
melaniebmn, line 102: Same comment about the target prop, can we change that to only have it displayed if it's not set to its default value?

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
melaniebmn, line 88: Can we change the display prop to the following to only have it displayed if it's not set to its default value?
```${args.display != 'inline' ? `display="${args.display}"` : null}```

File: packages/web/src/components/gcds-link/stories/gcds-link.stories.js [hunk 50]
```diff
@@ -0,0 +1,211 @@
...
+  default: 'an external',
+};
+
+export const Download = Template.bind({});
+Download.args = {
+  href: 'test.jpg',
+  target: '_self',
+  size: 'regular',
+  download: 'file.pdf',
+  type: 'image/jpg',
```
melaniebmn, line 180: I'm sorry I feel really picky but can we change the value for `type` to `application/pdf` since the file type is a pdf and not an image?

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,211 @@
...
+    args.external ? `external` : null
+  } ${args.download ? `download="${args.download}"` : null} ${
+    args.type ? `type="${args.type}"` : null
+  }>${args.default}</GcdsLink>
+link.
+`.replace(/ null/g, '');
+
+const TemplateSizes = () => `
+<!-- Web component code (Angular, Vue) -->
+<gcds-link href="" size="regular">This is a regular link.</gcds-link>
```
melaniebmn, line 113: Can we remove `size="regular"` since `regular` is the default value for the `size prop, users don't have to add this.

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,211 @@
...
+<gcds-link href="" size="small">This is a small link.</gcds-link>
+
+<!-- React code -->
+<GcdsLink href="" size="regular">This is a regular link</GcdsLink>
```
melaniebmn, line 117: Can we remove `size="regular"` since `regular` is the default value for the `size prop, users don't have to add this.

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,211 @@
...
+    args.target ? `target="${args.target}"` : null
+  } ${args.size != 'regular' && args.size ? `size="${args.size}"` : null} ${
+    args.external ? `external` : null
+  } ${args.download ? `download="${args.download}"` : null} ${
+    args.type ? `type="${args.type}"` : null
+  }>${args.default}</GcdsLink>
+link.
+`.replace(/ null/g, '');
+
+const TemplateSizes = () => `
```
melaniebmn, line 111: Can we add an example for a link with the `inherit size` as well? Maybe it can be inside an h2 or something to make the size different more obvious.

packages/web/src/components/gcds-link/stories/overview.mdx
```diff
@@ -0,0 +1,63 @@
+import { Meta, Canvas, Story } from '@storybook/blocks';
+import * as Link from './gcds-link.stories';
+
+<Meta of={Link} name="Overview" />
+
+# Link<br/>`<gcds-link>`
+
+_Also called: call to action, CTA, link._
```
melaniebmn, line 8: This seems to still be using the button content. Can we change this to:
`Also called: anchor, hyperlink, call to action`

packages/web/src/components/gcds-link/stories/overview.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+A Link is an interactive object that highlights an important or common action.
```
melaniebmn, line 10: This seems to still be using the button content. Can we change this to:
`A navigational element that allows users to navigate to a new page, website, or section on the current page.`

packages/web/src/components/gcds-link/gcds-link.css
```diff
@@ -0,0 +1,30 @@
...
+   font: var(--gcds-link-small-font);
+  }
+}
```
melaniebmn, line 30: Just realized:

Since we are using a slot in this component, we need to use the slot bug fix that Ethan worked on this week:
```
slot {
    display: inherit;
  }
  ```

File: packages/web/src/components/gcds-link/stories/gcds-link.stories.js [hunk 57]
```diff
@@ -0,0 +1,211 @@
...
+} ${args.download ? `download="${args.download}"` : null} ${
+  args.type ? `type="${args.type}"` : null
+}>  ${args.default}
+</gcds-link>
+`;
+
+// ------ Link Default ------
+export const Default = Template.bind({});
+Default.args = {
+  display: '',
```
melaniebmn, line 137: Can you set `display` to `inline` everywhere it's using the default value? It ensures that the properties table will load the correct values.

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,211 @@
...
+  href: '#',
+  rel: '',
+  target: '',
```
melaniebmn, line 140: Can you set `target` to `_self` everywhere it's using the default value? It ensures that the properties table will load the correct values.

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,211 @@
...
+};
+
+// ------ Link events & props ------
+export const Props = Template.bind({});
+Props.args = {
+  display: '',
+  href: '#',
+  rel: '',
+  target: '',
+  size: '',
```
melaniebmn, line 155: Can you set `size` to `regular` everywhere it's using the default value? It ensures that the properties table will load the correct values.

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,211 @@
...
+<gcds-link ${args.display ? `display="${args.display}"` : null} ${
+  args.href ? `href="${args.href}"` : null
+} ${args.rel ? `rel="${args.rel}"` : null} ${
+  args.target ? `target="${args.target}"` : null
+} ${args.size != 'regular' && args.size ? `size="${args.size}"` : null} ${
+  args.external ? `external` : null
+} ${args.download ? `download="${args.download}"` : null} ${
+  args.type ? `type="${args.type}"` : null
+}>  ${args.default}
+</gcds-link>
```
melaniebmn, line 131: Is this link displaying for you? It doesn't actually display for me when I go to the `Playground` tab.

melaniebmn (changes requested):
Aded some more comments, please let me know if you have any questions!

packages/web/src/components/gcds-link/gcds-link.css
daine: I think we have a lot of other components that don't have this, I will add it here, but maybe something we can add to the backlog to update? I just did a quick search for everything that has `&:focus`

ethanWallace: A lot of other components have the border radius applied before focus but a good example of this is in the [details component](https://github.com/cds-snc/gcds-components/blob/develop/packages/web/src/components/gcds-details/gcds-details.css#L27) (that's my goto component when I need to remember focus styles)

packages/web/src/components/gcds-link/gcds-link.css
daine: Good cach

packages/web/src/components/gcds-link/gcds-link.css
daine: Awesome, thanks Ethan!

packages/web/src/components/gcds-link/gcds-link.tsx
daine: Yep -- good catch on the default settings for the display!

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
daine: Not picky at all! This is just random typing, I'll make sure it's consistent for sure, even in the href.

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
daine: I had to set a value for href (I set it to '#') because it's a required property

packages/web/src/components/gcds-link/i18n/i18n.js
```diff
@@ -0,0 +1,16 @@
...
+    email: 'Opens an email client.',
+  },
+  fr: {
+    external: "S'ouvre dans un nouvel onglet.",
+    phone: 'Phone number [fr & copy edit needed]',
+    download: 'Download [fr & copy edit needed]',
+    email: 'Email [fr & copy edit needed]',
+  },
+};
+
```
daine, line 15: Awaiting Amy's translation and approval

packages/web/src/components/gcds-link/i18n/i18n.js
SmartMouthWords: Thanks for the discussion earlier this week @daine.  
 
 
    external: 'Opens destination in a new tab.',
    phone: 'Attempts to open a phone app.',
    download: 'Attempts to download a file to the device.',
    email: 'Opens new message in email reader.',
    email: 'Opens new message in email reader.',
OR

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,221 @@
...
+  type: '',
+  default: 'a GCDS Link component',
+};
+
+export const External = Template.bind({});
+External.args = {
+  display: 'inline',
+  href: 'http://design-system.alpha.canada.ca',
+  rel: '',
+  target: '_self',
```
melaniebmn, line 169: Can we set this to `_blank` since this is an example of an external link?

packages/web/src/components/gcds-link/gcds-link.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+    if (!values.includes(newValue)) {
+      this.size = 'inherit';
+    }
+  }
+
+  /**
+   * Sets the display behavior of the link
+   */
+  @Prop() display?: 'block' | 'inline' | 'inline-block' = 'inline';
```
melaniebmn, line 46: Sorry, I'm gonna be annoying again :D I just remembered that CSS has some weird specs where it doesn't actually show the `text-decoration` for inline blocks. Since we want to use underlines for links to clearly mark them as links, let's remove the `inline-block` option everywhere, please.

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,221 @@
...
+    args.external ? `external` : null
+  } ${args.download ? `download="${args.download}"` : null} ${
+    args.type ? `type="${args.type}"` : null
+  }>${args.default}</GcdsLink>
+link.
+`.replace(/ null/g, '');
+
+const TemplateSizes = () => `
+<!-- Web component code (Angular, Vue) -->
+<gcds-link href="#" size="regular">This is a regular link.</gcds-link>
```
melaniebmn, line 113: Can we remove the `size="regular"` part here since `regular` is the default value for the `size` prop, users don't have to add this.

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,221 @@
...
+<gcds-link href="#" size="small">This is a small link.</gcds-link>
+<gcds-heading tag="h2"><gcds-link href="#">This is a link with size set to inherit (default).</gcds-link></gcds-heading>
+
+<!-- React code -->
+<GcdsLink href="#" size="regular">This is a regular link</GcdsLink>
```
melaniebmn, line 118: Can we remove the `size="regular"` part here since `regular` is the default value for the `size` prop, users don't have to add this.

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,221 @@
...
+    args.target != '_self' ? `target="${args.target}"` : null
+  } ${args.size != 'inherit' && args.size ? `size="${args.size}"` : null} ${
+    args.external ? `external` : null
+  } ${args.download ? `download="${args.download}"` : null} ${
+    args.type ? `type="${args.type}"` : null
+  }>${args.default}</GcdsLink>
+link.
+`.replace(/ null/g, '');
+
+const TemplateSizes = () => `
```
melaniebmn, line 111: Looking at the overview page now, it's a bit hard to see the different size options. Would you mind adding a subheading under size + a story for each of the 3 size options? Thank you!!!

packages/web/src/components/gcds-link/gcds-link.css
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  &:focus {
+    border-radius: var(--gcds-link-focus-border-radius);
+    background-color: var(--gcds-link-focus-background);
+    color: var(--gcds-link-focus-text);
+    box-shadow: var(--gcds-link-focus-box-shadow);
+    outline: var(--gcds-link-focus-outline-width) solid var(--gcds-link-focus-background);
+    outline-offset: var(--gcds-link-focus-outline-offset);
+    border-color: var(--gcds-link-focus-border-color);
```
melaniebmn, line 24: Apologies for not catching this when you created the token but since the border colour + background colour are supposed to be the same colour, you can use the `--gcds-link-focus-background` token here instead of creating a separate token for the focus border. Can you please remove that token from the token repo? Again sorry for not catching this in the token PR!

melaniebmn (changes requested):
Almost there!!! Only a few more comments and then this should be good to ship!
Something I noticed on the Storybook page:

When the `display` prop is set to `block` and `external` is set to `true`, the link text and the icon get displayed on separate lines. Can you look into this and ensure they get displayed on the same line?

Here is a picture of it:

Screenshot 2023-11-06 at 5 07 45 PM

packages/web/src/components/gcds-link/gcds-link.tsx
daine: Not annoying; good catch!

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
daine: I think in a previous review you said the default size would be "inherit"?

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
daine: I think in a previous review you said the default size would be "inherit"?

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
daine: Yup I can put that again and we can keep whichever looks best for sure

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
melaniebmn: Oh yes, that is my bad - you are absolutely right!

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
melaniebmn: You are right, please ignore this.

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
melaniebmn: Sorry, I meant to add a different story for each in the overview tab. I can add a suggestion for it.

packages/web/package.json
```diff
@@ -43,7 +43,7 @@
     "@babel/core": "^7.20.12",
     "@babel/preset-env": "^7.20.2",
     "@babel/preset-typescript": "^7.21.0",
-    "@cdssnc/gcds-tokens": "^1.9.2",
+    "@cdssnc/gcds-tokens": "^1.10.0",
```
melaniebmn, line 46: The token version will need to be updated to the latest version.

packages/web/src/components/gcds-link/stories/overview.mdx
```diff
@@ -0,0 +1,63 @@
...
+  of={Link.Phone}
+  story={{ inline: true }}
+/>
+
+
+### Sizes
+<Canvas
+  of={Link.TemplateSizesExample}
+  story={{ inline: true }}
+/>
```
melaniebmn, line 49:
```suggestion
### Sizes

#### Size inherit
<Canvas
  of={Link.TemplateSizesInherit}
  story={{ inline: true }}
/>

#### Size regular
<Canvas
  of={Link.TemplateSizesRegular}
  story={{ inline: true }}
/>

#### Size small
<Canvas
  of={Link.TemplateSizesSmall}
  story={{ inline: true }}
/>
```

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,221 @@
...
+<!-- Web component code (Angular, Vue) -->
+<gcds-link href="#" size="regular">This is a regular link.</gcds-link>
+<gcds-link href="#" size="small">This is a small link.</gcds-link>
+<gcds-heading tag="h2"><gcds-link href="#">This is a link with size set to inherit (default).</gcds-link></gcds-heading>
+
+<!-- React code -->
+<GcdsLink href="#" size="regular">This is a regular link</GcdsLink>
+<GcdsLink href="#" size="small">This is a small link</GcdsLink>
+<GcdsHeading tag="h2"><GcdsLink href="#">This is a link with size set to inherit (default).</GcdsLink></GcdsHeading>
+`;
```
melaniebmn, line 121:
```suggestion
Remove TemplateSizes
```

melaniebmn, line 121:
```suggestion
Remove TemplateSizes
```

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,221 @@
...
+export const Phone = Template.bind({});
+Phone.args = {
+  display: 'inline',
+  target: '_self',
+  size: 'inherit',
+  href: 'tel:1234567890',
+  default: 'a phone number',
+};
+
+export const TemplateSizesExample = TemplateSizes.bind({});
```
melaniebmn, line 206:
```suggestion
export const SizesInherit = Template.bind({});
SizesInherit.args = {
   display: 'inline',
  href: '#',
  rel: '',
  target: '_self',
  size: 'inherit',
  external: false,
  download: '',
  type: '',
  default: 'link size inherit',
};

export const SizesRegular= Template.bind({});
SizesRegular.args = {
   display: 'inline',
  href: '#',
  rel: '',
  target: '_self',
  size: 'regular',
  external: false,
  download: '',
  type: '',
  default: 'link size regular',
};

export const SizesSmall= Template.bind({});
SizesSmall.args = {
   display: 'inline',
  href: '#',
  rel: '',
  target: '_self',
  size: 'small',
  external: false,
  download: '',
  type: '',
  default: 'link size regular',
};
```

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,246 @@
...
+  target: '_self',
+  size: 'small',
+  external: false,
+  download: '',
+  type: '',
+  default: 'link size small',
+};
+
+export const SizesRegular = Template.bind({});
+Default.args = {
```
melaniebmn, line 208: Copy & paste error on my end - this should be:
`SizesRegular.args = {`

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,246 @@
...
+  target: '_self',
+  size: 'regular',
+  external: false,
+  download: '',
+  type: '',
+  default: 'link size regular',
+};
+
+export const SizesInherit = Template.bind({});
+Default.args = {
```
melaniebmn, line 221: Copy & paste error on my end - this should be:
`SizesInherit.args = {`

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,246 @@
...
+Phone.args = {
+  display: 'inline',
+  target: '_self',
+  size: 'inherit',
+  href: 'tel:1234567890',
+  default: 'a phone number',
+};
+
+export const SizesSmall = Template.bind({});
+Default.args = {
```
melaniebmn, line 195: Copy & paste error on my end - this should be:
`SizesSmall.args = {`

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,246 @@
...
+  } ${args.rel ? `rel="${args.rel}"` : null} ${
+    args.target != '_self' ? `target="${args.target}"` : null
+  } ${args.size != 'inherit' && args.size ? `size="${args.size}"` : null} ${
+    args.external ? `external` : null
+  } ${args.download ? `download="${args.download}"` : null} ${
+    args.type ? `type="${args.type}"` : null
+  }>${args.default}</GcdsLink>
+link.
+`.replace(/ null/g, '');
+
```
melaniebmn, line 110: Forgot in my earlier suggestion that the `inherit` link was supposed to be wrapped in a heading. We need to add a new template for that:

```suggestion
const TemplateSizeInherit = args =>
  `
<!-- Web component code (Angular, Vue) -->
<gcds-heading tag="h4"><gcds-link href="#">${args.default}</gcds-link></gcds-heading>

<!-- React code -->
<GcdsHeading tag="h4"><GcdsLink href="#">${args.default}</GcdsLink></GcdsHeading>
`;
```

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
```diff
@@ -0,0 +1,246 @@
...
+  rel: '',
+  target: '_self',
+  size: 'regular',
+  external: false,
+  download: '',
+  type: '',
+  default: 'link size regular',
+};
+
+export const SizesInherit = Template.bind({});
```
melaniebmn, line 220: Replace `Template` with `TemplateSizeInherit`

melaniebmn (changes requested):
Sooooo close! We still need to update the token version and then I added a few comments to fix my earlier suggestions - apologies about that, I made the suggestions without testing them :/

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
daine: Yeah, absolutely! I had the idea yesterday just didn't write the code haha thanks Melanie!

packages/web/src/components/gcds-link/stories/gcds-link.stories.js
daine:
```suggestion
export const SizesInherit = TemplateSizeInherit.bind({});
```

packages/web/src/components/gcds-link/gcds-link.css
```diff
@@ -0,0 +1,47 @@
+:host a {
+  cursor: pointer;
+  color: var(--gcds-link-default);
+
+  &:not([aria-disabled='true']) {
+    text-decoration: underline;
+    text-underline-offset: var(--gcds-link-underline-offset);
+    text-decoration-color: currentColor;
+    text-decoration-thickness: var(--gcds-link-decoration-thickness);
+  }
```
melaniebmn, line 10: I just noticed that we still had the check for disabled in here as a leftover from the button css. Let's remove it since it's not needed. Suggesting to change line 2-10 to this:

```suggestion
 cursor: pointer;
 color: var(--gcds-link-default);
 text-decoration: underline;
 text-underline-offset: var(--gcds-link-underline-offset);
 text-decoration-color: currentColor;
 text-decoration-thickness: var(--gcds-link-decoration-thickness);
```

packages/web/src/components/gcds-link/stories/overview.mdx
```diff
@@ -0,0 +1,77 @@
...
+  of={Link.SizesSmall}
+  story={{ inline: true }}
+/>
+
+
+## Resources
+
+<ul>
+  <li>
+    <gcds-link href="https://design-system.alpha.canada.ca/en/components/link/" target="_blank">Guidance</gcds-link>
```
melaniebmn, line 69: Noticed another small thing: can we set all the resources links to `size="regular"` please?

```suggestion
    <gcds-link size="regular" href="https://design-system.alpha.canada.ca/en/components/link/" target="_blank">Guidance</gcds-link>
```

packages/web/src/components/gcds-link/stories/overview.mdx
```diff
@@ -0,0 +1,77 @@
...
+  </li>
+  <li>
+    <gcds-link href="https://github.com/cds-snc/gcds-components/tree/main/packages/web/src/components/gcds-link" target="_blank">Github</gcds-link>
+  </li>
+  <li>
+    <gcds-link href="https://www.figma.com/file/mh2maMG2NBtk41k1O1UGHV/Canadian-Digital-Service%E2%80%A8---GC-Design-System?node-id=817-4607&t=CvcydrJWvNLwlAvY-0" target="_blank">Figma</gcds-link>
```
melaniebmn, line 75: Same comment about the size.

melaniebmn (changes requested):
A couple more comments but I swear that's gonna be it after that :D

packages/web/src/components/gcds-link/gcds-link.css
daine: Good catch!

melaniebmn (approved):
SHIP IT!!!!!!!! WOHOOOOO

PR #306, opened 2023-10-10
Add heading component

melaniebmn:
# Summary | Résumé

Add new gcds-heading component, introducing desktop and mobile typography styles for headings.

### Tag
Sets the appropriate HTML tag for the selected level.

### Character limit
Sets the line length to a maximum amount of characters per line for each heading level, ensuring a comfortable, accessible reading length.

### Margin-top
Adds margin above the heading.

### Margin-bottom
Adds margin below the heading.

Note: I will add the gcds-heading component to our components together with the gcds-text component in a new PR.


ethanWallace (dismissed):
LGTM!

melaniebmn:
@ethanWallace reverting this back to a draft since we decided on adding the default margins.

packages/web/src/components/gcds-heading/stories/gcds-heading.stories.tsx
```diff
@@ -0,0 +1,190 @@
...
+        '550',
+        '600',
+        '700',
+        '800',
+        '900',
+        '1000',
+      ],
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: 'h1: 0, h2-h6: 500' },
```
melaniebmn, line 48: Does this look ok as default value or is it weird to display the different values?

melaniebmn:
Update: Added the default margin values for the heading levels. Margin-bottom will use `400` as default for all heading levels. H1 will use `0` as the default value for margin-top, while the h2-h6 headings will use `500` as the default value.

packages/web/src/components/gcds-heading/stories/gcds-heading.stories.tsx
ethanWallace: I really like this solution!

Are we going to also display the default value for the margin-bottom in storybook?

melaniebmn: We sure are - good catch! I just added it in.

ethanWallace (approved):
LGTM!

PR #311, opened 2023-10-19
add gcds-text component

melaniebmn:
# Summary | Résumé

Add new `gcds-text` component.

### Character limit
Sets the line length to a maximum amount of characters per line to ensure a comfortable, accessible reading length.

### Display
Specifies the display behavior of the text.

### Margin-top
Adds margin above the text.

### Margin-bottom
Adds margin below the text.

### Text-role
Sets the main style of the text.

### Size
Sets the appropriate HTML tags for the selected size.

### Style
Indicates if the text should be styled italic or normal. 

### Weight
Determines the font weight for the text.

### DIFF
--- a/packages/web/.storybook/preview-head.html
+++ b/packages/web/.storybook/preview-head.html
@@ -2,13 +2,18 @@
 <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
 <link
   href="https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,100;0,300;0,400;0,700;0,900;1,100;1,300;1,400;1,700;1,900&display=swap"
-  rel="stylesheet" />
+  rel="stylesheet"
+/>
 <link
   href="https://fonts.googleapis.com/css2?family=Noto+Sans:ital,wght@0,300;0,400;0,600;0,700;0,800;1,300;1,400;1,600;1,700;1,800&display=swap"
-  rel="stylesheet" />
+  rel="stylesheet"
+/>
 
 <!-- Font Awesome (Icons) -->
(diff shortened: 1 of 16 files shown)


packages/web/src/components/gcds-text/gcds-text.css
```diff
@@ -0,0 +1,248 @@
...
+  :host .gcds-text {
+    font: var(--gcds-text-size-body-desktop);
+
+    @media only screen and (width < 48em) {
+      font: var(--gcds-text-size-body-mobile);
+    }
+  }
+}
+
+@layer variants.display {
```
ethanWallace, line 33: Small question: For these display variants to apply, wouldn't they need to be applied to the host element? Since the host element has `display:block`, changing the display property on the `p` tag wouldn't change the how the text displays.

ethanWallace (commented):
Just a couple comments/questions

packages/web/src/components/gcds-text/gcds-text.css
melaniebmn: Good catch, I updated it!

packages/web/.storybook/preview-head.html
```diff
@@ -259,11 +267,17 @@
     background-color: var(--gcds-color-grayscale-50);
   }
 
+  /* Code preview white space */
+  .language-html {
+    white-space: break-spaces !important;
+    border: 1px solid red;
```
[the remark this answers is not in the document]
melaniebmn: Oops, forgot to remove that line - fixed it.

ethanWallace (approved):
LGTM!

PR #315, opened 2023-10-23
Add screen reader only component

ethanWallace:
# Summary | Résumé

Add `gcds-sr-only` component to component library.

## Usage

To use the screen reader only component, simply wrap the text you want hidden in the component

```html
<gcds-sr-only>
  Text only available for assistive technologies.
</gcds-sr-only>
```

### DIFF
--- a/packages/angular/src/lib/stencil-generated/components.ts
+++ b/packages/angular/src/lib/stencil-generated/components.ts
@@ -861,6 +861,27 @@ export declare interface GcdsSignature extends Components.GcdsSignature {}
 
 
 @ProxyCmp({
+})
+@Component({
+  selector: 'gcds-sr-only',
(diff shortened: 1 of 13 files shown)


melaniebmn (dismissed):
Looks great! The only suggestion I have: could you also add this component to the index.html file? I know it won't actually show anything but it's more to keep an inventory of what we have.

Feel free to merge without it though if you disagree.

ethanWallace:
> Looks great! The only suggestion I have: could you also add this component to the index.html file? I know it won't actually show anything but it's more to keep an inventory of what we have.
> 
> Feel free to merge without it though if you disagree.

Good catch! Thought I included it already but I clearly missed it!

melaniebmn (approved):
LGTM - thanks for adding!

PR #316, opened 2023-10-25
remove text-style and weight props from gcds-text to avoid misuse

melaniebmn:
# Summary | Résumé

Removing `text-style` and `weight` props from the `gcds-text` component to avoid misuse. Including them as props and adding these styles to an entire block of text would lead to a bad and less accessible user experience and therefore they should not be available as default props.


daine (approved):
LGTM! (Could probably use another review from @ethanWallace)

ethanWallace (approved):
LGTM

PR #318, opened 2023-10-27
Add slot styles to help with chrome accessibility tree rendering

ethanWallace:
# Summary | Résumé

To fix the accessibility tree not populating properly in some versions of Chrome, add some default slot styles into component CSS. This was added to any component that could possibly receive static text through a slot, when text is passed in an element it seems to handle it properly.


daine (approved):
LGTM!

melaniebmn (approved):
LGTM - can we document this somewhere, please? Or at least have a card on the dev board to document this?

PR #325, opened 2023-10-31
fix: Change slot CSS in gcds-breadcrumbs to display text underline

ethanWallace:
# Summary | Résumé

With the new slot CSS added to the components, the `gcds-breadcrumbs-items` component no longer had a text underline. Modified the slot CSS to display the underline properly.


melaniebmn (approved):
LGTM

PR #334, opened 2023-11-08
update demo page to use gcds-heading component

melaniebmn:
# Summary | Résumé

Update the demo page to use the gcds-heading component instead of plain heading tags.


ethanWallace (approved):
LGTM

PR #337, opened 2023-11-14
fix gcds-link external icon spacing

melaniebmn:
# Summary | Résumé

The external icon for the link component was using `margin-left="200"` instead of `margin-left="100"` like the other icons for this component. Fixed the value to `100` to be consistent.


ethanWallace (approved):
LGTM!

PR #338, opened 2023-11-14
Improve storybook a11y

melaniebmn:
# Summary | Résumé

In our last round of Fable testing, some users pointed out that it was really hard to see where the focus was when interacting with our live demo control elements. I updated the focus to be much more visible.

The users also pointed out that it was difficult/impossible to use the control toggles as they aren't coded accessibly by Storybook and therefore the labels don't get read aloud properly. After testing Storybook's radio buttons, I realized that those weren't coded properly either. Because of the poor accessibility of toggles and radio buttons within Storybook, I switched all toggles and radio buttons in our stories to use selects as this seems to be the only accessible option available.

### DIFF
--- a/packages/web/src/components/gcds-error-summary/stories/gcds-error-summary.stories.tsx
+++ b/packages/web/src/components/gcds-error-summary/stories/gcds-error-summary.stories.tsx
@@ -37,7 +37,7 @@ const Template = args =>
   `
 <!-- Web component code (Angular, Vue) -->
 <gcds-error-summary
-  ${args.listen && args.errorLinks ? `listen` : null}
+  ${args.listen && !args.errorLinks ? `listen` : null}
   ${args.errorLinks ? `error-links='${args.errorLinks}'` : null}
   ${args.heading ? `heading="${args.heading}"` : null}
 >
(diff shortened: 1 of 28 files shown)


packages/web/src/components/gcds-error-summary/stories/gcds-error-summary.stories.tsx
```diff
@@ -36,15 +37,15 @@ const Template = args =>
   `
 <!-- Web component code (Angular, Vue) -->
 <gcds-error-summary
-  ${args.listen && !args.errorLinks ? `listen` : null}
+  ${args.listen && args.errorLinks ? `listen` : null}
```
ethanWallace, line 40: This logic is to set either `listen` or `error-links` will need to be set back to `${args.listen && !args.errorLinks ? `listen` : null}`

packages/web/src/components/gcds-error-summary/stories/gcds-error-summary.stories.tsx
```diff
@@ -36,15 +37,15 @@ const Template = args =>
...
   ${args.errorLinks ? `error-links='${args.errorLinks}'` : null}
   ${args.heading ? `heading="${args.heading}"` : null}
 >
 </gcds-error-summary>
 
 <!-- React code -->
 <GcdsErrorSummary
-  ${args.listen && !args.errorLinks ? `listen` : null}
```
ethanWallace, line 47: Same as above

ethanWallace (changes requested):
Just one small thing with `gcds-error-summary` code rendering

ethanWallace (approved):
LGTM

PR #346, opened 2023-11-24
Update stencil configuration for better build

ethanWallace:
# Summary | Résumé

Update stencil to `v4.7.2` and the update and remove some configuration properties in `stencil.config.ts`.

- Added `validatePrimaryPackageOutputTarget` to validate package on build
- Remove `buildEs5` as it was causing issues with some frameworks
- Updated `experimentalImportInjection` to `enableImportInjection` as it is no longer experimental
- Removed `"sourceMaps": true` from `tsconfig` in react package to hopefully stop warnings when using components in `create-react-app`. This will also decrease the package size.

### DIFF
--- a/packages/react/tsconfig.json
+++ b/packages/react/tsconfig.json
@@ -16,7 +16,6 @@
     "outDir": "dist",
     "removeComments": false,
     "skipLibCheck": true,
-    "sourceMap": true,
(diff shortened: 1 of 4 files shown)


melaniebmn (approved):
LGTM

PR #350, opened 2023-11-30
fix: change slot styles to initial

melaniebmn:
# Summary | Résumé

We added slot styles a few weeks ago to fix a Chrome bug that kept the accessibility tree from populating properly. Setting the slot style to `inherit` was causing some issues with links not displaying the underline in a few places. Changing the slot styles to `initial` to prevent that issue.

The linter also added a few formatting updates :)


ethanWallace (approved):
LGTM!

Idk if we want to wait to merge this after the `v0.17.1` release work is sorted out or try to include this in that release.

PR #363, opened 2023-12-21
fix: link variant prop name

melaniebmn:
# Summary | Résumé

Rename `link-variant` prop to `variant` to be consistent with other components.


daine (approved):
LGTM!

PR #364, opened 2023-12-21
fix: File-uploader not validating properly

ethanWallace:
# Summary | Résumé

Fix issue where the `gcds-file-uploader` would prevent form submission and display an error after choosing a file when validating `onBlur`. Added a small timeout to the validation in `handleChange` function to allow the file-uploader component to re-render with the submitted file.


daine (approved):
LGTM

melaniebmn (approved):
LGTM

PR #367, opened 2024-01-10
Add wrapped prop for conditional render to fix slot hydration issue

ethanWallace:
# Summary | Résumé

When using the experimental SSR package, a hydration error would occur when a GCDS component was slotted into another GCDS component. This was happening since the `GcdsWrapper` wraps each  GCDS component when used causing conflicts in the render.  This can be seen with the breadcrumbs, top-nav and side-nav components using https://github.com/ethanWallace/components-ssr-test.

## Solution

Added logic to add a `wrapped` prop into the rendering to prevent a `GcdsWrapper` from being placed inside another `GcdsWrapper`. The new prop is eventually removed before final render so it doesn't display on the component. These props should only be passed to GCDS components and avoid any other component.

### DIFF
--- a/packages/ssr/scripts/utils.ts
+++ b/packages/ssr/scripts/utils.ts
@@ -78,6 +78,10 @@ const ${toPascalCase(elementName)}WebComponent = React.forwardRef(({ children = 
     nativeProps[p.replace(/([a-z0-9])([A-Z])/g, '$1-$2').toLowerCase()] = nonEventProps[p];
   }
 
+  if (nativeProps.wrapped) {
+    delete nativeProps.wrapped;
+  }
+
   if (typeof window !== 'undefined') {
     const innerRef = useRef();
     useImperativeHandle(ref, () => innerRef.current);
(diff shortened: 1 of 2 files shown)


daine:
I'm for this fix - appreciate the work put in!
Do we know if there are any side effects of this? Does it take longer to render?

ethanWallace:
> Do we know if there are any side effects of this? Does it take longer to render?

I honestly can't say. Before the developer would have to manually wrap our components in `GcdsWrapper` but we are trying to just build that in and I can't say if the `GcdsWrapper` would cause any issues with other libraries.

daine:
What do you think about doing `wrap` instead of `wrapped`? 

I like it because it's simple easy to understand -- `wrapped` seems to mean "this is already wrapped so **_do not_** wrap it" vs `wrap` which means "wrap this".

packages/ssr/scripts/utils.ts
```diff
@@ -90,10 +94,42 @@ const ${toPascalCase(elementName)}WebComponent = React.forwardRef(({ children =
 
 function ${toPascalCase(elementName)}(props) {
   if (!props['noSSR'] && typeof window == 'undefined') {
+    let cloned = [];
+
+    // Pass wrapped prop to children GCDS components
+    if (Array.isArray(props.children)) {
+      for (let x = 0; x< props.children.length; x++) {
```
melaniebmn, line 101: You could simplify the structure a little bit if you use a map function for the array iteration instead.

melaniebmn (changes requested):
One code improvement suggestion, but otherwise looks good. Thanks for looking into this and working on a fix!

ethanWallace:
> what do you think about doing `wrap` instead of `wrapped`?
> 
> I like it because it's simple easy to understand -- `wrapped` seems to mean "this is already wrapped so **_do not_** wrap it" vs `wrap` which means "wrap this".

Wrapped is the more accurate name. The new prop is only applied to the component and it's children when a component is rendered with `GcdsWrapper` so it is wrapped and now uses that prop to prevent rendering another `GcdsWrapper` inside the first one.

daine:
> > what do you think about doing `wrap` instead of `wrapped`?
> > I like it because it's simple easy to understand -- `wrapped` seems to mean "this is already wrapped so **_do not_** wrap it" vs `wrap` which means "wrap this".
> 
> Wrapped is the more accurate name. The new prop is only applied to the component and it's children when a component is rendered with `GcdsWrapper` so it is wrapped and now uses that prop to prevent rendering another `GcdsWrapper` inside the first one.

I was attempting to mean the opposite - using property names that require a detailed explanation tends to introduce more cognitive load and bugs to a codebase which is the reason why I'm not a big fan of them. If we do `wrap` it would do the exact same thing (wrap it in `GcdsWrapper` because it is not yet wrapped, otherwise do not add a `GcdsWrapper`

ethanWallace:
I think we are splitting hairs by trying to figure out if we do the logic based off a true prop or a false prop. I do see issues with switching the logic up like how does the first element know to be wrapped but I can definitely switch the prop to a clearer name.

daine:
> I think we are splitting hairs by trying to figure out if we do the logic based off a true prop or a false prop. I do see issues with switching the logic up like how does the first element know to be wrapped but I can definitely switch the prop to a clearer name.

Thank you, let's give it a clearer name and hopefully we can remove the comments too for the self-explanatory parts once that's done!

daine (changes requested):
Could we change `wrapped` to be more clear and concise?

Example:
```javascript
if( prop.isAlreadyWrapped ){
        <${toPascalCase(elementName)}WebComponent {...props}>
        {cloned}
        </${toPascalCase(elementName)}WebComponent>
}
return {
   <GcdsWrapper>
        <${toPascalCase(elementName)}WebComponent {...props}>
        {cloned}
        </${toPascalCase(elementName)}WebComponent>
   </GcdsWrapper>
}
```

daine (approved):
LGTM! Thanks for making the change

melaniebmn (approved):
LGTM

PR #370, opened 2024-01-10
fix: update renovate base branch

melaniebmn:
Switching Renovate to use `main` as the base branch instead of `develop` since we will be removing the `develop` branch very soon.


daine (approved):
Great catch, thank you and LGTM!

PR #376, opened 2024-01-18
Form-associated components

ethanWallace:
# Summary | Résumé

Update all form components to use shadow-dom and take advantage of new form-associated feature.

## Changes
- Changed the way `gcds-select` populates the options, it still has options passed by the `slot` but the previous way wouldn’t work in shadow-dom. (May not be SSR friendly but will look at that later)
- Updated fieldset validation

More info on [form-associated components in Stencil](https://stenciljs.com/docs/form-associated)

### DIFF
--- a/packages/ssr/lib/client/index.ts
+++ b/packages/ssr/lib/client/index.ts
@@ -36,10 +36,21 @@ function resolve(children, result = []) {
         resolve(vnode, result);
       } else {
         // Function component
-        resolve(node.type(node.props), result);
+        const props = { ...node.props };
+        // Apply extra isAlreadyWrapped prop to GCDS components
+        if (node.type.toString().includes('Gcds')) {
+          props.isAlreadyWrapped = true;
+        }
+        resolve(node.type(props), result);
       }
     } else if (typeof node.type === 'object' && typeof node.type.render === 'function') {
-      resolve(node.type.render(node.props), result);
+      const props = { ...node.props };
+      const renderedComp = node.type.render(node.props);
+      // Apply extra isAlreadyWrapped attribute to GCDS components
+      if (renderedComp.type.includes('gcds-')) {
+        props.isAlreadyWrapped = true;
+      }
+      resolve(node.type.render(props), result);
     }
   }
 
--- a/packages/ssr/scripts/utils.ts
+++ b/packages/ssr/scripts/utils.ts
@@ -78,6 +78,10 @@ const ${toPascalCase(elementName)}WebComponent = React.forwardRef(({ children = 
     nativeProps[p.replace(/([a-z0-9])([A-Z])/g, '$1-$2').toLowerCase()] = nonEventProps[p];
   }
 
+  if (nativeProps.isAlreadyWrapped) {
+    delete nativeProps.isAlreadyWrapped;
+  }
+
   if (typeof window !== 'undefined') {
     const innerRef = useRef();
     useImperativeHandle(ref, () => innerRef.current);
@@ -90,10 +94,41 @@ const ${toPascalCase(elementName)}WebComponent = React.forwardRef(({ children = 
 
 function ${toPascalCase(elementName)}(props) {
   if (!props['noSSR'] && typeof window == 'undefined') {
+    let cloned = [];
+
+    // Pass isAlreadyWrapped prop to children GCDS components
+    if (Array.isArray(props.children)) {
+      props.children.map((child, i) => {
+        if (child.type && child.type.toString().includes('Gcds')) {
+          cloned.push(React.cloneElement(child, { isAlreadyWrapped: true, key: i }));
+        } else if (typeof child == 'string') {
+          cloned.push(child);
+        } else {
+          cloned.push(React.cloneElement(child, { key: i }));
+        }
+      });
+    } else if (typeof props.children == 'object') {
+      if (props.children.type && props.children.type.toString().includes('Gcds')) {
+        cloned.push(React.cloneElement(props.children, { isAlreadyWrapped: true, key: 0 }));
+      } else {
+        cloned.push(React.cloneElement(props.children, { key: 0 }));
+      }
+    } else if (typeof props.children == 'string') {
+      cloned = props.children;
+    }
+
+    if (props.isAlreadyWrapped) {
+      return(
+        <${toPascalCase(elementName)}WebComponent {...props}>
+          {cloned}
+        </${toPascalCase(elementName)}WebComponent>
+      )
+    }
+
     return(
       <GcdsWrapper>
         <${toPascalCase(elementName)}WebComponent {...props}>
-          {props.children}
+          {cloned}
         </${toPascalCase(elementName)}WebComponent>
       </GcdsWrapper>
     )
--- a/packages/web/src/components/gcds-checkbox/gcds-checkbox.css
+++ b/packages/web/src/components/gcds-checkbox/gcds-checkbox.css
@@ -1,129 +1,129 @@
-:host {
-  .gcds-checkbox {
+@layer reset, default, disabled, error, focus;
+
+@layer reset {
+  :host {
+    display: block;
+
+    .gcds-checkbox {
+      padding: 0;
+
+      gcds-label {
+        &:before,
+        &:after {
+          cursor: pointer;
+          content: '';
+          box-sizing: border-box;
+        }
+      }
+    }
+  }
+}
+
+@layer default {
+  :host .gcds-checkbox {
     max-width: 90%;
     font: var(--gcds-checkbox-font);
     color: var(--gcds-checkbox-default-text);
-    margin: var(--gcds-checkbox-margin);
-    padding: 0;
+    margin: var(--gcds-checkbox-margin) !important;
     transition: color ease-in-out 0.15s;
     position: relative;
 
-    &:focus-within {
-      color: var(--gcds-checkbox-focus-text);
-
-      gcds-hint {
-        color: currentColor;
-      }
+    :is(gcds-label, gcds-hint, gcds-error-message) {
+      padding: var(--gcds-checkbox-label-padding) !important;
     }
 
-    &.gcds-checkbox--disabled {
-      color: var(--gcds-checkbox-disabled-text);
-
-      input[type=checkbox] + gcds-label {
-        color: var(--gcds-checkbox-disabled-text);
-
-        &:before {
-          border-color: currentcolor;
-          background-color: var(--gcds-checkbox-disabled-background);
-          cursor: not-allowed;
-        }
-
-        &:after {
-          cursor: not-allowed;
-        }
-      }
-
-      gcds-hint {
-        color: currentColor;
-      }
-    }
-
-    &.gcds-checkbox--error:not(:focus-within) input[type=checkbox] + gcds-label {
-      &:before {
-        border-color: var(--gcds-checkbox-danger-border);
-      }
-
-      &:after {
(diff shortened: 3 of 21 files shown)


packages/web/src/components/gcds-fieldset/gcds-fieldset.tsx
```diff
@@ -29,8 +29,7 @@ import i18n from './i18n/i18n';
 @Component({
   tag: 'gcds-fieldset',
   styleUrl: 'gcds-fieldset.css',
-  shadow: false,
-  scoped: true,
+  shadow: { delegatesFocus: true },
```
daine, line 32: Do we need `formAssociated: true,` here? Just ignore if we don't

packages/web/src/components/gcds-input/gcds-input.tsx
```diff
@@ -248,15 +240,31 @@ export class GcdsInput {
...
 
-  handleChange(e) {
-    if (this.changeHandler) {
-      this.changeHandler(e);
-    } else {
-      const val = e.target && e.target.value;
-      this.value = val;
+  @Listen('keydown', { target: 'document' })
+  keyDownListener(e) {
+    if (e.target == this.el && e.key == 'Enter') {
```
daine, line 245: small suggestion:
```suggestion
    if (e.target == this.el && e.key === 'Enter') {
```

packages/web/src/components/gcds-input/test/gcds-input.e2e.ts
daine, line 1: Do we have a test for the "Enter" keypress?

packages/web/src/components/gcds-select/gcds-select.tsx
```diff
@@ -180,13 +177,9 @@ export class GcdsSelect {
...
 
-  private onBlur = e => {
-    if (this.focusHandler) {
-      this.focusHandler(e);
-    } else {
-      if (this.validateOn == 'blur') {
-        this.validate();
-      }
+  private onBlur = () => {
+    if (this.validateOn == 'blur') {
```
daine, line 181: Could we take this opportunity to do a strict comparison?
```suggestion
    if (this.validateOn === 'blur') {
```

packages/web/src/components/gcds-select/gcds-select.tsx
```diff
@@ -274,6 +279,22 @@ export class GcdsSelect {
...
+          this.value = this.options[x].getAttribute('value');
+          this.initialValue = this.value ? this.value : null;
+        }
+
+        if (this.value == this.options[x].getAttribute('value')) {
+          this.options[x].setAttribute('selected', 'true');
+          this.internals.setFormValue(this.options[x].getAttribute('value'));
+          this.initialValue = this.value;
+        }
+      }
```
daine, line 296: Could we use map instead?
```suggestion
      this.options.map((option) => {
        if (option.hasAttribute('selected')) {
          this.value = option.getAttribute('value');
          this.initialValue = this.value ? this.value : null;
        }

        if (this.value === option.getAttribute('value')) {
          option.setAttribute('selected', 'true');
          this.internals.setFormValue(option.getAttribute('value'));
          this.initialValue = this.value;
        }
      });
```

daine (changes requested):
Looking great! Just a few things I noticed

packages/web/src/components/gcds-fieldset/gcds-fieldset.tsx
ethanWallace: No we won't since fieldset only groups form elements and doesn't submit anything

daine (approved):
LGTM!

packages/web/src/components/gcds-error-summary/gcds-error-summary.tsx
```diff
@@ -205,13 +207,17 @@ export class GcdsErrorSummary {
               Object.keys(errorQueue).map(key => {
                 return (
                   <li class="summary__listitem">
-                    <a
-                      onClick={e => this.focusElement(e, key)}
-                      class="summary__link"
-                      href={key}
+                    <gcds-link
```
melaniebmn, line 210: Since you are replacing the `a.summary__link` with the `gcds-link` component, can you also please remove the CSS for `.summary__link` and those tokens?

packages/web/src/components/gcds-error-summary/gcds-error-summary.tsx
```diff
@@ -76,15 +78,17 @@ export class GcdsErrorSummary {
...
 
   @Listen('gcdsValid', { target: 'document' })
   validListener(e) {
     if (this.listen && e.target.closest('form') == this.el.closest('form')) {
-      if (this.errorLinksObject && this.errorLinksObject[e.detail.id]) {
-        delete this.errorLinksObject[e.detail.id];
+      for (const [key, value] of Object.entries(this.errorLinksObject)) {
+        if (value == e.target) {
+          delete this.errorLinksObject[key];
+        }
```
melaniebmn, line 91: Instead of looping through the object's entries to delete keys, could you use `filter` to create a new object without the keys you want to remove? This would make it a bit more functional and easier to read.

packages/web/src/components/gcds-error-summary/gcds-error-summary.tsx
```diff
@@ -119,11 +123,11 @@ export class GcdsErrorSummary {
...
     const sortable = [];
-    for (const id in this.errorLinksObject) {
+    for (const [key, value] of Object.entries(this.errorLinksObject)) {
       sortable.push([
-        id,
-        this.errorLinksObject[id],
-        document.querySelector(id).getBoundingClientRect().y,
+        key,
+        value,
+        (value as HTMLElement).getBoundingClientRect().y,
```
melaniebmn, line 130: Could you use `map` instead of `forEach` to create a new object that is a little bit more concise?
Instead of manually creating an object with `forEach`, you can use `Object.fromEntries` to create the object from an array of key-value pairs.

Picturing something like this (full disclosure - didn't test this):

```
sortErrors() {
  const sortable = Object.entries(this.errorLinksObject).map(([key, value]) => [
    key,
    value,
    (value as HTMLElement).getBoundingClientRect().y,
  ]);

  sortable.sort((a, b) => a[2] - b[2]);

  return Object.fromEntries(sortable);
}
```

packages/web/src/components/gcds-file-uploader/gcds-file-uploader.tsx
```diff
@@ -199,24 +185,22 @@ export class GcdsFileUploader {
...
-      const files = e.target.files;
+    const filesContainer: string[] = [];
+    const files = e.target.files;
 
-      for (let i = 0; i < files.length; i++) {
-        filesContainer.push(files[i].name);
-      }
+    for (let i = 0; i < files.length; i++) {
+      filesContainer.push(files[i].name);
+    }
```
melaniebmn, line 193: Could you use `map` instead of `for` to iterate over the files to simplify the code?

packages/web/src/components/gcds-file-uploader/gcds-file-uploader.tsx
```diff
@@ -286,6 +280,32 @@ export class GcdsFileUploader {
...
+
+  /*
+   * Set form data for internals
+   */
+  private addFilesToFormData = files => {
+    const formData = new FormData();
+
+    for (let i = 0; i < files.length; i++) {
+      formData.append(this.name, files[i], files[i].name);
+    }
```
melaniebmn, line 304: If you use `forEach` here instead of `for`, the code will be a bit easier to read.
Something like this:
```
Array.from(files).forEach(file => {
    formData.append(this.name, file, file.name);
});
  ```

File: packages/web/src/components/gcds-select/gcds-select.tsx [hunk 10]
```diff
@@ -274,6 +279,21 @@ export class GcdsSelect {
     }
 
     this.inheritedAttributes = inheritAttributes(this.el, this.shadowElement);
+
+    if (this.el.children) {
+      this.options = Array.from(this.el.children);
+      this.options.map(option => {
```
melaniebmn, line 285: Since you are not mapping to a new array, using `forEach` here would be a bit more efficient.

packages/web/src/components/gcds-select/gcds-select.tsx
```diff
@@ -274,6 +279,21 @@ export class GcdsSelect {
     }
 
     this.inheritedAttributes = inheritAttributes(this.el, this.shadowElement);
+
+    if (this.el.children) {
+      this.options = Array.from(this.el.children);
```
melaniebmn, line 284: Could we specifically target the option elements and be more concise by doing something like this?
`const options = Array.from(this.el.querySelectorAll('option'));`

packages/web/src/components/gcds-select/gcds-select.tsx
melaniebmn: Additionally, you can reduce some code repetition by creating a variable for `.getAttribute('value')`:
`const value = option.getAttribute('value');`

packages/web/src/components/gcds-select/gcds-select.tsx
```diff
@@ -338,20 +361,28 @@ export class GcdsSelect {
...
           >
             {defaultValue ? (
               <option value="" disabled selected>
                 {defaultValue}
               </option>
             ) : null}
-
-            <slot></slot>
+            {options.map(opt => {
+              const selected = {};
```
melaniebmn, line 376: Instead of adding an empty object and conditionally adding properties to it, you can use a ternary operator to conditionally add the `selected` attribute.

`const selected = opt.hasAttribute('selected') ? { selected: true } : null;`

packages/web/src/components/gcds-select/gcds-select.tsx
```diff
@@ -338,20 +361,28 @@ export class GcdsSelect {
...
+              if (opt.hasAttribute('selected')) {
+                selected['selected'] = true;
+              }
```
melaniebmn, line 379: Instead of adding an empty object and conditionally adding properties to it, you can use a ternary operator to conditionally add the `selected` attribute.

`const selected = opt.hasAttribute('selected') ? { selected: true } : null;`

melaniebmn (changes requested):
Added a few suggestions, but aside from that - awesome work! Thank you so much for taking this one!

packages/web/src/components/gcds-select/gcds-select.tsx
ethanWallace: I used `this.el.children` to just not focus on `options` but the rest of my code just assumed the passed elements were `<options>` so it doesn't make sense to use `this.el.children`. But I just remembered `<optgroup>` are a thing so I am going to refactor the rendering to allow them to work with `<optgroup>`'s

melaniebmn: Gotcha, that makes sense - thank you!

packages/web/src/components/gcds-error-summary/gcds-error-summary.tsx
ethanWallace: Since it is an object, there doesn't seem to be an easy `filter` function for it and most other ways I have found online seem to do it this way

melaniebmn (approved):
There is one merge conflict but PR itself looks good to go!

Please also add a PR to the tokens repo to remove those `summary__link` tokens that are not needed anymore or create a ticket in Zenhub so we don't loose track of it.

melaniebmn (approved):
Ship it!

PR #381, opened 2024-01-23
Add additional check for href property in gcds-link

ethanWallace:
# Summary | Résumé

Add additional check for `href` property to solve error warning in Next.js. The error would always seem to occur event when an `href` value was present.


daine (approved):
LGTM!

PR #388, opened 2024-01-30
feat: lerna monorepo config

daine:
# Summary | Résumé
While working on `release-please`, it was having issues picking up the structure of the monorepo and updating the package versions.

This change will fully utilize `lerna` and `npm` [workspaces](https://docs.npmjs.com/cli/v10/using-npm/workspaces) for a monorepo repository, and will make it easier for `release-please` to handle.

**_I will create a separate PR for the `release-please` work_**

Part of [667](https://github.com/cds-snc/design-gc-conception/issues/667)

Few things to note:
```javascript
# The order of the packages is important
  "workspaces": [
    "packages/web",
    "packages/angular",
    "packages/react"
  ],
```

With workspaces, `npm install` takes care of installing all the packages so there is no need to run it on every child package. For the most part, running scripts will also not require you to `cd` into `packages/web`. Example:
```bash
// on root directory
$ npm run start

> @cdssnc/gcds-components@0.18.1 start
> stencil build --dev --watch --serve
```

Future suggestions:
- We may want to add tests for our wrapper packages

⚠️ ~~I will hold off merge until I'm confident `compile-and-publish` will work as expected~~ **Tested the workflow without actually publishing it and it runs smoothly so far**

Check that build files are similar to published packages
- [x] [React package](https://www.npmjs.com/package/@cdssnc/gcds-components-react?activeTab=code)
- [x] [Angular package](https://www.npmjs.com/package/@cdssnc/gcds-components-angular?activeTab=code)
- [x] [Web components package](https://www.npmjs.com/package/@cdssnc/gcds-components?activeTab=code)

### DIFF
--- a/.github/workflows/build-components.yml
+++ b/.github/workflows/build-components.yml
@@ -1,18 +1,12 @@
-name: Renovate pull request
+name: Build components
 on:
   pull_request:
       branches:
         - main
-      types:
-        - labeled
-        - opened
-        - synchronize
-        - reopened
 
 jobs:
   build-packages:
     name: Build all packages
-    if: contains(github.event.pull_request.labels.*.name, 'Renovate')
     runs-on: ubuntu-latest
     steps:
       - name: Git Checkout
@@ -63,7 +57,6 @@ jobs:
 
   build-storybook:
     name: Build storybook
-    if: contains(github.event.pull_request.labels.*.name, 'Renovate')
     runs-on: ubuntu-latest
     steps:
       - name: Git Checkout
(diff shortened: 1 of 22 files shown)


.github/workflows/build-components.yml
```diff
@@ -21,39 +21,9 @@ jobs:
 
       - name: Install gcds-components
```
ethanWallace, line 22: You can remove the `install gcds-components` step since the previous step(`install repo`) also runs `npm install`.

ethanWallace (changes requested):
Just one small thing but this is great!

.github/workflows/build-components.yml
daine: Oh yes great catch!

melaniebmn (approved):
Ship it - wohooooo!!!

ethanWallace (approved):
Ship it!

PR #391, opened 2024-01-31
refactor: css rewrite label

melaniebmn:
# Summary | Résumé

CSS rewrite for the label component to add new CSS features like layers, etc.

Part of fix for https://github.com/cds-snc/design-gc-conception/issues/627

### DIFF
--- a/packages/web/src/components/gcds-label/gcds-label.css
+++ b/packages/web/src/components/gcds-label/gcds-label.css
@@ -1,24 +1,33 @@
-:host .gcds-label {
-  display: block;
-  max-width: 100%;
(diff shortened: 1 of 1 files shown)


daine (commented):
Was the `!important` added because the change to `@layers` made it lower priority?

Confirming we don't need `cursor: pointer` for `gcds-checkbox` and `gcds-radio` labels anymore?

daine (approved):

melaniebmn:
@daine I added `cursor: pointer;` as default styling to the label so there is no more need to specify it only for the checkbox and radio components.

PR #394, opened 2024-01-31
refactor: css rewrite icon

melaniebmn:
# Summary | Résumé

CSS rewrite for the icon component to add new CSS features like layers, etc.


ethanWallace:
When running this change with `npm run start` the icons appear normally, but if I run this change in `storybook` the icons only appear as a box.

After some playing around I was able to make them appear when switching to use an `import` of the font awesome css in a new layer.

melaniebmn:
Thanks for catching this and also finding a fix!!! I'm hoping to be able to work on the new icon component in the next sprint or latest the one after and then we can luckily get rid of Font Awesome and won't run into these issues anymore.

PR #415, opened 2024-02-01
refactor: gcds-radio to gcds-radio-group

ethanWallace:
# Summary | Résumé

Refactor `gcds-radio` into `gcds-radio-group` to allow native browser controls over radio buttons.

## Changes

Removed a variety of properties and moved them into an `object` to pass to the new `options` attribute.

## Example

```html
<gcds-radio
  name="" 
  options=’[{
    “id”: “”,
    “label”: “”,
    “required”: “”,
    “disabled”: “”,
    “value”: “”,
    “checked”: “”,
    “hint”: “”,
  }]’
>
</gcds-radio>
```


melaniebmn (commented):
Code looks good - 1 idea/question though: can we include the fieldset directly into this component to save the user from having to add it in?

ethanWallace:
> Code looks good - 1 idea/question though: can we include the fieldset directly into this component to save the user from having to add it in?

I was thinking of do that but didn't end up doing it since I wanted the `gcds-fieldset`'s use to be consistent with other form components. But with that being said I could bring this to dev/design sync and see what people think about this idea.

melaniebmn:
> > Code looks good - 1 idea/question though: can we include the fieldset directly into this component to save the user from having to add it in?
> 
> I was thinking of do that but didn't end up doing it since I wanted the `gcds-fieldset`'s use to be consistent with other form components. But with that being said I could bring this to dev/design sync and see what people think about this idea.

That makes sense! Good idea to bring it to dev-design sync to see how the majority feels about it from a UX experience. :)

daine (approved):
LGTM

melaniebmn (approved):
LGTM

PR #425, opened 2024-02-05
refactor: css rewrite top-nav

melaniebmn:
# Summary | Résumé

CSS rewrite for the top-nav component to add new CSS features like layers, etc.

Part of fix for https://github.com/cds-snc/design-gc-conception/issues/627.

Note: Had to add a wrapper div to be able to use container queries.

### DIFF
--- a/packages/web/src/components/gcds-top-nav/gcds-top-nav.css
+++ b/packages/web/src/components/gcds-top-nav/gcds-top-nav.css
@@ -1,4 +1,4 @@
-@layer reset, default, compact;
+@layer reset, default, desktop;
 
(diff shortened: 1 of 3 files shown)


ethanWallace (changes requested):
Seeing a couple weird things with this one.

In storybook, the component seems to not render properly. It has no grey background and almost seems to render like the sidebar.

I also noticed the mobile menu(after you click the large menu button) no longer becomes an overlay. The menu just opens where it is placed on the page

ethanWallace (approved):
LGTM!

PR #426, opened 2024-02-07
refactor: css rewrite side-nav

melaniebmn:
# Summary | Résumé

CSS rewrite for the side-nav component to add new CSS features like layers, etc.

Part of fix for https://github.com/cds-snc/design-gc-conception/issues/627.

### DIFF
--- a/packages/web/src/components/gcds-side-nav/gcds-side-nav.css
+++ b/packages/web/src/components/gcds-side-nav/gcds-side-nav.css
@@ -44,10 +44,6 @@
         padding: 0;
         overflow: hidden;
       }
-
-      gcds-nav-group {
-        padding: 0;
-      }
     }
   }
 }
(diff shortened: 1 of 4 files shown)


packages/web/src/components/gcds-side-nav/gcds-side-nav.css
```diff
@@ -1,34 +1,53 @@
...
+      .gcds-side-nav__heading {
         display: block;
         width: 0;
         height: 0;
         margin: 0;
         padding: 0;
         overflow: hidden;
       }
+
+      gcds-nav-group {
```
ethanWallace, line 48: I noticed this line of `padding: 0` on the `gcds-nav-group` is causing the close button in the mobile overlay to push right up against the edges of the container. When it receives focus the focus outline is cut off on either side making it look a little weird.

ethanWallace (approved):
LGTM!

PR #427, opened 2024-02-08
fix: storybook spacing in component preview

melaniebmn:
# Summary | Résumé

A recent Storybook update changed the class name for one of our styles. I updated it to add the styling to a different (semantic) class in the preview code which Storybook normally doesn't update. If they change it again, I will change it to use the tag, but normally they do stick to their semantic class names.

Before:
Screenshot 2024-02-08 at 11 48 02 AM

After:
Screenshot 2024-02-08 at 11 47 54 AM

### DIFF
--- a/packages/web/.storybook/preview-head.html
+++ b/packages/web/.storybook/preview-head.html
@@ -45,9 +45,9 @@
     border: none !important;
   }
 
-  /* Copy & hide code btns */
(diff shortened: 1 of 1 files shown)


ethanWallace (approved):
LGTM!

daine (approved):

PR #434, opened 2024-02-15
ci: add workflow dispatch to release generator gh action

daine:
# Summary | Résumé
I had the release please workflow paused and it didn't run, and realized it can't be manually dispatched.

This PR adds the `workflow_dispath` trigger to release please so we can choose to run it manually if we need to.


ethanWallace (approved):
LGTM

PR #436, opened 2024-02-15
ci: workflow fixes

daine:
# Summary | Résumé
GH Action for release please behaves differently from testing on local environment. I'll test the workflow on this PR instead to get a better view of what it's trying to do.

Things changed:
- The Release Please GH Action changed a few things (which we want!), the biggest thing is that we had the setting `release-type: node` set which seemed innocuous but [it actually now stops release please from using the manifest file](https://github.com/google-github-actions/release-please-action?tab=readme-ov-file#upgrading-from-v3-to-v4)
- Release please config for `node-workspaces` plugin is needed to update the peer dependencies. I thought about taking it out because it shows up as additional lines in the generated CHANGELOG, but without it, the PR fails due to the peer dependencies not being updated to the new version.

This [PR](https://github.com/cds-snc/gcds-components/pull/440) is an example of the release pr created by these changes. Once this is merged, I expect it to simply update https://github.com/cds-snc/gcds-components/pull/439

### DIFF
--- a/.github/workflows/release-generator.yml
+++ b/.github/workflows/release-generator.yml
@@ -27,4 +27,3 @@ jobs:
         with:
           token: ${{ steps.sre_app_token.outputs.token }}
           release-type: node
-          include-component-in-tag: true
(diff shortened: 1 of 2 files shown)


ethanWallace (approved):
LGTM

PR #443, opened 2024-02-16
fix: Storybook events and property page accessibility

ethanWallace:
# Summary | Résumé

Add script to document head to run various fixes for storybook events and properties page. The script will only run when the query strings `demo="true"` and `lang="en | fr"` are present.

Example url: `/iframe.html?viewMode=docs&demo=true&lang=fr&singleStory=true&id=components-input--events-properties`

## Fixes

- https://github.com/cds-snc/design-gc-conception/issues/694
- https://github.com/cds-snc/design-gc-conception/issues/691
- https://github.com/cds-snc/design-gc-conception/issues/687
- https://github.com/cds-snc/design-gc-conception/issues/686
- Tagged english only content, like property names, with `lang="en"`

### DIFF
--- a/packages/web/.storybook/preview-head.html
+++ b/packages/web/.storybook/preview-head.html
@@ -118,22 +118,6 @@
     color: var(--gcds-text-primary) !important;
   }
 
-  .docblock-argstable-head th:nth-child(1) span:before {
-    content: 'Property | Propriété';
-  }
-
-  .docblock-argstable-head th:nth-child(2) span:before {
-    content: 'Data type | Type de donnée';
-  }
-
-  .docblock-argstable-head th:nth-child(3) span:before {
-    content: 'Default value | Valeur par défaut';
-  }
-
-  .docblock-argstable-head th:nth-child(4) span:before {
-    content: 'Values | Valeurs';
-  }
-
   /* Table Styling */
   .docblock-argstable-head tr th,
   .docblock-argstable-body tr td.css-11nj37o,
--- a/packages/web/.storybook/preview-head.html
+++ b/packages/web/.storybook/preview-head.html
@@ -391,6 +391,8 @@
 	      formatTable(document.querySelector('.docblock-argstable'), url.get('lang'));
       }
     }, 200);
+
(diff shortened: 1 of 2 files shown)


daine:
http://localhost:6006/iframe.html?viewMode=docs&demo=true&lang=fr&singleStory=true&id=components-input--events-properties
Screenshot 2024-02-16 at 9 37 01 AM

http://localhost:6006/iframe.html?viewMode=docs&demo=true&lang=en&singleStory=true&id=components-input--events-properties
Screenshot 2024-02-16 at 9 37 32 AM

packages/web/.storybook/preview-head.html
```diff
@@ -374,4 +359,169 @@
...
+        }
+        // Format third cell
+        if (tr.children[x].innerText === '-') {
+          if (lang === 'en') {
+            tr.children[x].setAttribute('title', 'No default value');
+          } else {
+            tr.children[x].setAttribute('title', 'Aucune valeur par défaut');
+            tr.children[x].setAttribute('lang', 'fr');
+          }
+        }
```
daine, line 504: I can see the new "No default value" added as a td title, but I can't find it on the a11y tree nor when I attempt to use voiceover.

Storybook Embed: In the code builder for all components, rows in the Properties table without a default value indicate that there is no default using a hyphen. This may be clear for sighted users but a text alternative should be provided for screen reader and other assistive technology users (e.g. “No default value”).

ethanWallace: Right it seems the extra logic I had for that isn't here. Will push it up soon

melaniebmn (changes requested):
There is a drop-shadow left on the table body now which makes it look a little odd. Can you remove that as well, please? Adding this should fix it:

```
.css-quknxc.css-quknxc tbody {
    box-shadow: none !important;
    filter: none !important;
}
```

melaniebmn (approved):
Ship it 👍

daine (approved):
Thank you, this is great! LGTM

PR #455, opened 2024-02-27
Add workflow dispatch to build-components.yml

daine:
Adding `workflow_dispatch:` so we can run this action manually, and also run this on the ssr branch automatically as well.


ethanWallace (approved):
LGTM

PR #456, opened 2024-02-28
refactor: form-associated components + gcds-radio-group rework

ethanWallace:
# Summary | Résumé

**Documentation will need to be updated before this can be merged**

Bring component work that has been done in the `gcds-components-ssr` branch into `main` to publish before `@cdssnc/gcds-components-ssr` is ready to publish.

## Changes
- Changed the way `gcds-select` populates the options, it still has options passed by the `slot` but the previous way wouldn’t work in shadow-dom.
- Updated fieldset validation
- First phase of CSS rewrite
- Refactor `gcds-radio` into `gcds-radio-group`

### gcds-radio-group example

```html
<gcds-radio-group
  name="" 
  options='[{
    "id": "",
    "label": "",
    "required": "",
    "disabled": "",
    "value": "",
    "checked": "",
    "hint": "",
  }]'
>
</gcds-radio-group>
```


melaniebmn (approved):
LGTM, awesome work

daine (changes requested):
Looks great! We're going to need to flag this as a breaking change, one thing we can do right now is to change the PR title to have a `!` after the type/scope like this:
> refactor!: form-associated components + gcds-radio-group rework

...which will get picked up as the commit message before/during squash + merge. This will immediately signal it as a breaking change and will update the CHANGELOGs as such.

ethanWallace:
Closing as we will launch this work with the SSR package

PR #457, opened 2024-02-28
refactor: integrate gcds-text into existing components 

melaniebmn:
# Summary | Résumé

Integrate gcds-text component into existing components to reduce code duplication created through repeated text styles.

gcds-text was integrated into:

- gcds-card
- gcds-date-modified
- gcds-error-message
- gcds-file-uploader
- gcds-hint
- gcds-textarea

[Token PR](https://github.com/cds-snc/gcds-tokens/pull/232)

### DIFF
--- a/packages/web/src/components/gcds-card/gcds-card.css
+++ b/packages/web/src/components/gcds-card/gcds-card.css
@@ -36,18 +36,9 @@
       width: 100%;
     }
 
-    .gcds-card__tag {
-      font: var(--gcds-card-tag-font);
-      color: var(--gcds-card-tag-color);
-    }
-
     .gcds-card__title {
       font: var(--gcds-card-title-font);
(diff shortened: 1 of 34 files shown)


ethanWallace:
It looks good! Just have one question before giving it a review.

With these changes the hint no longer receives the focus colour. I know design has decided to remove the focus colour from label and hint ☹️ so is the hint no longer changing the expected behaviour?

melaniebmn:
> It looks good! Just have one question before giving it a review.
> 
> With these changes the hint no longer receives the focus colour. I know design has decided to remove the focus colour from label and hint ☹️ so is the hint no longer changing the expected behaviour?

Good catch @ethanWallace , fixed it!

ethanWallace (approved):
LGTM!

PR #468, opened 2024-03-11
fix: remove skip-to-content role from button

melaniebmn:
# Summary | Résumé

Remove the `skip-to-content` role from the button component now that we have the link component and have updated the header component to use the link for the skip-to-content link instead.


ethanWallace (approved):
LGTM

PR #470, opened 2024-03-13
feat: set allowed values for grid tag property to limit misuse

melaniebmn:
# Summary | Résumé

Limiting the allowed values for the grid tag property after some feedback we received from the OCADU report.

Linked to
- https://github.com/cds-snc/design-gc-conception/issues/757


daine (approved):
LGTM!

PR #471, opened 2024-03-13
fix: invalid path for types

### LINKED ISSUE #466: Missing type declarations in gcds-components-react
I'm getting the following error when using `gcds-components-react` with typescript in v0.19.0 and later:

`TS7016: Could not find a declaration file for module '@cdssnc/gcds-components-react'`

Let me know if I'm missing something, but I believe [this change](https://github.com/cds-snc/gcds-components/commit/424b138d6ebf8e3df77b2e265ff386bf1e39d7cf#diff-1f344ac391eeecc21ec0f01fb07430a47f4b80d20485c125447d54c33c4bbfc4L23) is the cause. When I install `gcds-components-react`, the file is still at `dist/index.d.ts` so `package.json` seems to be pointing to the wrong place now, and changing that back locally has fixed the issue for me.

daine:
# Summary | Résumé
- Fixes #466

Tested with a small create react app with typescript. I saw the exact same message as detailed in #466, and tested this updated package on that app. This fix should resolve the issue; I don't see anymore typescript errors during testing.

### DIFF
--- a/packages/react/tsconfig.json
+++ b/packages/react/tsconfig.json
@@ -13,18 +13,20 @@
     "noImplicitReturns": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
-    "outDir": "dist",
+    "outDir": "./dist",
     "removeComments": false,
     "skipLibCheck": true,
     "jsx": "react",
     "target": "es2015",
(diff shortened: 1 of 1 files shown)


melaniebmn (approved):
LGTM

PR #474, opened 2024-03-15
fix: Declarative shadow-dom error in Chrome

ethanWallace:
# Summary | Résumé

Prevent errors from appearing in Chrome by adjusting render logic.

## Changes

- Added a list of our components that are form-associated to add `shadowrootdelegatesfocus` to template tag in DSD
- Fixed render function from attaching multiple template tags to custom elements in components

### DIFF
--- a/packages/web/stencil.config.ts
+++ b/packages/web/stencil.config.ts
@@ -51,5 +51,6 @@ export const config: Config = {
   validatePrimaryPackageOutputTarget: true,
   extras: {
     enableImportInjection: true,
+    experimentalSlotFixes: true
   },
 };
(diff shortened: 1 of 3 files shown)


packages/web/stencil.config.ts
```diff
@@ -51,5 +51,6 @@ export const config: Config = {
   validatePrimaryPackageOutputTarget: true,
   extras: {
     enableImportInjection: true,
+    experimentalSlotFixes: true
```
daine, line 54: Is there a way for us to guard for things like these that are experimental?

daine (commented):
I wonder if we should start working on some e2e tests to help test and review as well?

packages/web/stencil.config.ts
ethanWallace: This addition doesn't directly fix the issue this PR addresses. I just decided to add it into our build to hopefully protect against any slot issues.

melaniebmn (approved):
LGTM

PR #476, opened 2024-03-19
fix: Hydration isses with Host element

ethanWallace:
# Summary | Résumé

Build a function to generate the appropriate attributes on the Host element of gcds components to prevent hydration issues between server and client.

Notes:
- `gcds-nav-link` still causes issues, will solve in another PR that also addresses some other issues with the navigation components
- `gcds-grid-col` still causes hydration issues in Chrome, will look into later


melaniebmn (approved):
LGTM

PR #479, opened 2024-03-19
chore(storybook): update frame background colour

melaniebmn:
# Summary | Résumé

Update storybook frame background colour.


ethanWallace (approved):
LGTM

PR #481, opened 2024-03-21
fix: Update render function to handle more than one custom element

ethanWallace:
# Summary | Résumé

Missed a use case when applying fix from before. Render function should now be able to handle a component that has more than one of the same custom elements in it's shadow root (ex: `gcds-link` in `gcds-footer`). Also adds the required `shadowrootdelegatesfocus` attribute to the `template` element if custom element is `delegateFocus:true`.


melaniebmn (approved):
LGTM - ship it

PR #486, opened 2024-03-27
feat: Component rewrites (form-associated, gcds-radio-group, CSS rewrites) 

ethanWallace:
# Summary | Résumé

‼️ Blocker: **Documentation will need to be updated before this can be merged**

Bring component work that has been done in the `gcds-components-ssr` branch into `main` to publish before `@cdssnc/gcds-components-react-ssr` is ready to publish.

### Form associated

- Changed the way `gcds-select` populates the options, it still has options passed by the `slot` but the previous way wouldn’t work in shadow-dom.
- Updated fieldset validation
- Refactor `gcds-radio` into `gcds-radio-group`

#### gcds-radio-group example

```html
<gcds-radio-group
  name="" 
  options='[{
    "id": "",
    "label": "",
    "required": "",
    "disabled": "",
    "value": "",
    "checked": "",
    "hint": "",
  }]'
>
</gcds-radio-group>
```

### CSS rewrite

- Update GC Design System components to use CSS layers
- Updated components to use `gcds-heading`, `gcds-text` to remove duplicate CSS from components


.github/workflows/build-components.yml
```diff
@@ -4,7 +4,6 @@ on:
   pull_request:
       branches:
         - main
-        - gcds-components-ssr
```
daine, line 7: Do we want to keep this for now so we can check the build status on PRs against the SSR branch, if we're releasing form components first?

ethanWallace: Good idea

melaniebmn (approved):
LGTM

melaniebmn (approved):
LGTM

daine (approved):
LGTM

PR #491, opened 2024-04-02
fix: package path not found for gcds.css

daine:
# Summary | Résumé
Importing css in my react file like
```javascript
import "@cdssnc/gcds-components-react-ssr/gcds.css";
```

gives me the following error:
`Module not found: Package path ./gcds.css is not exported from package <path>/@cdssnc/gcds-components-react-ssr (see exports field in node_modules/@cdssnc/gcds-components-react-ssr/package.json)`

This PR fixes the issue


ethanWallace (approved):
LGTM

PR #495, opened 2024-04-05
fix: Update utility functions for more usability

ethanWallace:
# Summary | Résumé

Update the `assignLanguage` and `inheritAttributes` to be more usable and functional with GC Design System components going forward.

## assignLanguage changes

Previously the `assignLanguage` would check if a `lang` attribute has been assigned to the host element of a component if needed. If not found, the function would then check the document element to get the page language for the component.

To better match how `lang` works in HTML, the component will now check it's ancestors all the way up to the document element for a `lang` attribute. This allows developers to tag sections of a page a different language from the default page language and have the component inherit the correct language.

## inheritAttributes changes

`inheritAttributes` was built to take data attributes assigned to form components and pass them down to the input rendered in each form component and remove the attribute from the host element. With the introduction of form-associated components and shadow-dom, data attributes will no longer be passed down to the rendered input since the component is now recognized as a form input in HTML forms.

### DIFF
--- a/packages/web/src/utils/utils.ts
+++ b/packages/web/src/utils/utils.ts
@@ -12,7 +12,7 @@ export const inheritAttributes = (
   // Check for any aria or data attributes
   for (let i = 0; i < el.attributes.length; i++) {
     const attr = el.attributes[i];
-    if (attr.name.includes('aria-') || attr.name.includes('data-')) {
+    if (attr.name.includes('aria-')) {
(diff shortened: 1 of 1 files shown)


melaniebmn (approved):
LGTM - nice change for the language assignment!

daine (approved):
That's a neat function `closestElement`, LGTM!

PR #499, opened 2024-04-10
chore: add run script

daine:
# Summary | Résumé
Makes it easer to `npm run start` from the root of the folder


melaniebmn (approved):
LGTM

PR #500, opened 2024-04-12
fix: Remove CSS causing card title text to disappear on hover/focus

ethanWallace:
# Summary | Résumé

Removed CSS from `<gcds-card type="link">` that caused the card title link text to display as the same colour as the background producing an accessibility error when the card was focused and hovered at the same time.


daine (approved):
LGTM

melaniebmn (commented):
Do we want to replace this with a different hover styling?

melaniebmn (approved):
LGTM

PR #503, opened 2024-04-17
refactor: change gcds-error-message and gcds-hint to use slot instead of prop

melaniebmn:
# Summary | Résumé

As discussed in dev sync:

- Change gcds-error-message to use slot instead of message prop
- Change gcds-hint to use slot instead of hint prop


ethanWallace (approved):
LGTM!

melaniebmn:
> @melaniebmn do we want to merge this to the `refactor/form-rewrite` branch or onto main, so we can release it earlier?

I think it makes sense to go out with the other form changes. What do you think @ethanWallace?

ethanWallace:
Yeah I agree with putting this in with our other form changes.

PR #513, opened 2024-05-02
fix: emit proper events onChange

ethanWallace:
# Summary | Résumé

Noticed with recent event refactor the `gcdsChange` event wasn't actually being emitted, the `gcdsInput` event was being emitted on the `onChange` event.


melaniebmn (approved):
LGTM

PR #514, opened 2024-05-02
fix(ssr): render issues

daine:
# Summary
_Fix a render issue on a node that's just a string_
Found a render issue when I'm using the PhaseBanner inside the Header component
```
 ⨯ TypeError: Cannot read properties of null (reading 'useContext')
    at resolve (../../gcds-components/packages/react-ssr/dist/esm/lib/client/index.js:43:31)
    at resolve (../../gcds-components/packages/react-ssr/dist/esm/lib/client/index.js:28:13)
    at resolve (../../gcds-components/packages/react-ssr/dist/esm/lib/client/index.js:43:13)
    at eval (../../gcds-components/packages/react-ssr/dist/esm/lib/client/index.js:7:385)
digest: "3024408193"
```
Also found a render issue if you attempted to add a component that uses forwardRefs like the Link component from NextJS. Didn't find a fix for it, but I added the conditional logic to handle it once we have the fix so it's easier to go back to.

### DIFF
--- a/packages/react-ssr/lib/client/index.ts
+++ b/packages/react-ssr/lib/client/index.ts
@@ -2,11 +2,18 @@
 
 import { PropsWithChildren } from 'react';
 
-// Must go in a client component
-// > Otherwise will error:
-// > Attempted to call the default export of ... from the server but it's on the client.
-// > It's not possible to invoke a client function from the server, it can only be rendered
-// > as a Component or passed to props of a Client Component.
+/**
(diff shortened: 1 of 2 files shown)


daine:
@ethanWallace I haven't looked into this in a while, I don't think I need it anymore, but could you double check? TIA!

ethanWallace:
I doesn't look like we will need this change after the latest fixes to the react SSR package

daine:
Closing this as it is no longer needed

PR #516, opened 2024-05-06
feat: Add/create @cdssnc/gcds-components-vue package

ethanWallace:
# Summary | Résumé

For easier integration with Vue 3, add the `@cdssnc/gcds-components-vue` package. This provides the GC Design System components within Vue component wrappers.


packages/vue/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+``` html
+<style src='@cdssnc/gcds-components-vue/gcds.css'>
+    /* global styles */
+</style> 
+```
+
+### Using Vite
+
+If using [Vite](https://vitejs.dev/), additional configuration will need to be added to `vite.config.ts` to prevent Vue from logging a warning about failing to resolve components.
```
daine, line 39:
```suggestion
If using [Vite](https://vitejs.dev/), additional configuration will need to be added to `vite.config.ts` to prevent Vue from logging a warning about failing to resolve components e.g. "Failed to resolve component: my-component").
```
could we add the example stencil gives in their docs too?

melaniebmn (approved):
LGTM - ship it!

daine (changes requested):
Just adds one more update to the README -- I'm aware there might be translation changes as well.

Also, I tested this using this [example app](https://github.com/cds-snc/gcds-examples/pull/4/files) and it works pretty well (granted I only really tested the select and header components but it was fun seeing it!)

melaniebmn (dismissed):
LGTM

ethanWallace:
Hopefully that last commit was the last one

daine (approved):
LGTM! Thanks for this! We should also get our installation instructions updated once we have this package released.

melaniebmn (approved):
LGTM

PR #523, opened 2024-05-10
fix: Emit change event from form components

ethanWallace:
# Summary | Résumé

To make sure developers have access to both native HTML events and our custom events, I put in some logic that will emit the `change` event again to get passed the shadow-dom.


daine:
Thanks Ethan! Once this is out, could we put back the code on our site and bring back the CSP, what do you think?

daine (approved):
LGTM

melaniebmn (approved):
LGTM - thanks for jumping on this so quickly!

PR #524, opened 2024-05-10
feat: Add dist-hyrdate-app output

ethanWallace:
# Summary | Résumé

Add the `dist-hydrate-script` output to web components package. This will allow developers the tools to be able to render the Stencil web components on the server side in different frameworks available from `@cdssnc/gcds-components/hydrate`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 dist/
 www/
 loader/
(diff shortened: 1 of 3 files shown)


melaniebmn (approved):
LGTM

PR #527, opened 2024-05-13
chore: update storybook to latest version

melaniebmn:
# Summary | Résumé

Updating Storybook to the latest version so we can merge Renovate's dependency update PR's again.
We are currently blocked from merging them because the Storybook update needs manual work before it can be merged.

[Zenhub ticket](https://app.zenhub.com/workspaces/design-system-6100624a19f4cf000e46e458/issues/gh/cds-snc/design-gc-conception/898)


packages/web/.storybook/main.js
```diff
@@ -23,3 +25,7 @@ const config = {
   },
 };
 export default config;
+
+function getAbsolutePath(value) {
+  return dirname(require.resolve(join(value, "package.json")));
+}
```
daine, line 31: Nice!

daine (commented):
It looks great; I will load my environment and test it tomorrow and will approve once I'm able to see it, thanks for this!

ethanWallace (changes requested):
Just found a couple small things that need to be updated but I am not able to point to them directly, luckily they are all focused into the `preview-head.html`.

- The code comments in the code preview need to be set to white. The CSS selector can just change to `.token.comment` [line 382](https://github.com/cds-snc/gcds-components/blob/update-storybook/packages/web/.storybook/preview-head.html#L382)
- The copy code button in the code preview CSS class has seem to changed so we will need to update the class name in the CSS and in the `copyButtonTimer` function

PR #540, opened 2024-05-24
chore(storybook): add custom copy code button to storybook code preview

melaniebmn:
# Summary | Résumé

Adding custom copy code buttons to our Storybook code preview - one for the web component code and one for the React component code. Users are never going to need to copy both web and React, this gives them the option to choose.

I looked into separating the web components and React code preview into two. It would involve having 2 separate iframes with two separate property tables for which we would have to display tabs (or something similar) on the live demo section on our code page. We could look into doing that when we redesign the site, but it seems out of scope as a quick fix since it would require writing new stories for all components, design input, etc.

[Zenhub ticket](https://app.zenhub.com/workspaces/design-system-6100624a19f4cf000e46e458/issues/gh/cds-snc/design-gc-conception/862)

### DIFF
--- a/packages/web/.storybook/preview-head.html
+++ b/packages/web/.storybook/preview-head.html
@@ -65,16 +65,13 @@
     left: 0 !important;
   }
 
-  .css-3s4ai0 {
-    padding-block-end: var(--gcds-spacing-550);
-  }
-
   .css-3s4ai0 .css-111a2cx {
     background-color: transparent !important;
   }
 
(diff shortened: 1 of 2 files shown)


ethanWallace (changes requested):
This is great! Just noticed one thing.

If you hide then show the code block, the copy buttons no longer appear on the page. You could add an event listener to the show/hide button to run the `createBtn` functions again.

ethanWallace (approved):
LGTM!

Issue #534, opened 2024-05-21
bug: <gcds-topic-menu> doesn't look the between FR and EN


gcharest:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/cds-snc/gcds-components/issues) that already report this problem, without success.

### GC Design System Components Package and Version

gcds-components@0.21.0

### Current Behavior

The collapsed French button for the <gcds-topic-menu> component is narrower and uses more vertical space than the English version.

<img width="619" alt="image" src="https://github.com/cds-snc/gcds-components/assets/1690085/eb5c1640-3669-4d04-866d-f0c78f41f8a2">

<img width="533" alt="image" src="https://github.com/cds-snc/gcds-components/assets/1690085/80d2241e-9971-4943-acdb-386ca6ecafb4">
 

### Expected Behavior

They should be the exact same width and height and the chevron should be on the same line.

### System Info

```shell
The behaviour is observed on:
- Chrome 125.0.6422.61
- Firefox 126.0 (64-bit)
- Safari 17.5

Using macOS 14.5 on Intel MacBook Pro
```

### Steps to Reproduce

1. Visit the links to the components : https://design-system.alpha.canada.ca/en/components/theme-and-topic-menu/design/
2. Scroll down to the menu button
3. Toggle language switch
4. Scroll down to the menu button

melaniebmn:
Hi Guillaume, thank you for the bug report!

We fixed the issue in this [PR](https://github.com/cds-snc/gcds-components/pull/541), and it will go live with our next package release. Thank you so much for reporting this!

gcharest:
Thanks @melaniebmn ! I'll have a look at v0.22.0

I can also confirm that the issue wasn't present on v0.20.0 as I've been testing on various versions to see when the difference happened.

melaniebmn:
Hey @gcharest, the bug was fixed yesterday with the version `0.22.1` release.
[Here](https://github.com/cds-snc/gcds-components/blob/main/CHANGELOG.md#0221-2024-05-27) is the changelog for the change.

Closing the issue for now, but please let us know if it's still happening for you after upgrading to the newest version.

PR #541, opened 2024-05-24
fix: misaligned of FR theme and topic menu button (bug)

melaniebmn:
# Summary | Résumé

The French theme and topic menu button was split into two lines because the host element of the screenreader-only component was set to `display: block;`. Fixed it and it should display on one line now, like the EN menu button.

[Bug report](https://github.com/cds-snc/gcds-components/issues/534)


ethanWallace (approved):
LGTM

daine (approved):
LGTM!

PR #544, opened 2024-05-28
chore(storybook): card property typos in Storybook

ethanWallace:
# Summary | Résumé

Some properties for the card component were not rendering in the proper format (`imgAlt` instead of `img-alt`) on the Events & properties page in Storybook.


melaniebmn (approved):
LGTM

PR #550, opened 2024-05-30
chore(storybook): Stop copy buttons rendering more than once 

ethanWallace:
# Summary | Résumé

On the events and properties page, stop the web component and react copy buttons from rendering another time when modifying properties.

**Note**: This solution only works for the events and properties page and is a little hacky on top of our hacky JS. Will solve the other issues in another PR, just wanted to get the user facing issue fixed first.


melaniebmn (approved):
LGTM, thanks for jumping in!

daine (approved):

PR #551, opened 2024-05-31
chore(storybook): Fix display of overview copy buttons

ethanWallace:
# Summary | Résumé

Fix issue in storybook overview pages where the copy code buttons would disappear for all previews if the 'Hide code' button was toggled.

### DIFF
--- a/packages/web/.storybook/preview-head.html
+++ b/packages/web/.storybook/preview-head.html
@@ -680,7 +680,7 @@
       createCopyBtn(toggleCode, stringReact, '', 'React', 'React');
     } else {
       // If not expanded, remove existing copy buttons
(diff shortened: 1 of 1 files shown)


daine (changes requested):
Looks like the pages (Overview, Events & Properties) share the same toggle -- switching between them causes issues. I'm still getting same issue (missing copy buttons when the code section is open)
Screenshot 2024-06-04 at 11 32 04 AM

daine (approved):
There is some slightly odd behavior when you switch between the Overview & Events and Properties page, but it doesn't cause usability issues. It's a big improvement, thanks @ethanWallace! LGTM

PR #559, opened 2024-06-17
chore(deps): Update package dependencies to remove patch

ethanWallace:
# Summary | Résumé

Update `@cdssnc/gcds-components-react-ssr` to use an already patched version of stencil so we don't have to patch it ourselves. Will still need to apply patch to `@stencil/core` for running unit tests.


daine (changes requested):
_Adding some notes here for now, these are just some of my initial thoughts:_

I'm looking into the [patched package]( https://github.com/smartive/stencil-patched), and its corresponding [tags](https://github.com/smartive/stencil-patched/tags).

The latest tag `4.18.3` points to [this commit](https://github.com/smartive/stencil-patched/commit/0649166f542a5fd531e8d5e2f3e870af8095728a) which doesn't seem to have any similarities to the StencilJS [release for the same version v4.18.3](https://github.com/ionic-team/stencil/releases/tag/v4.18.3)?

Would be nice to have this piece clarified a bit more

ethanWallace:
We decided this was not the best approach to solve this issue. Will be looking into more options.

PR #570, opened 2024-06-25
refactor: Patched stencil dependency for React SSR package

ethanWallace:
# Summary | Résumé

To ensure the `@cdssnc/gcds-components-react-ssr` package has the required patch to the `@stencil/core` dependency, make a local version of the required file on build and patch it.

## Background

To achieve the SSR method used in the `@cdssnc/gcds-components-react-ssr` package, the file `@stencil/core/internal/client/index.js` needs to be updated to also include `globalThis`.

## Changes

- Updated `@stencil/core` to `v4.18.3`
- Removed need for running `patch-package` before `npm run build` on monorepo
- Added copying files and rewriting the imports in `dist` files for the `react-ssr` package on build


melaniebmn (approved):
LGTM

daine:
FYI I'd still like to test this -- hoping to get to this tomorrow :)

packages/react-ssr/scripts/build.ts
```diff
@@ -3,6 +3,7 @@ import { createRequire } from 'node:module';
 import { dirname, format, join, parse } from 'node:path';
 import { exit } from 'node:process';
 import { fileURLToPath } from 'node:url';
+import {replaceInFile} from 'replace-in-file'
```
daine, line 6: nit: lint fix
```suggestion
import { replaceInFile } from 'replace-in-file'
```

packages/react-ssr/package.json
```diff
@@ -43,7 +43,7 @@
     "gcds.css"
   ],
   "scripts": {
-    "prebuild": "del-cli --force ./dist && copyfiles -f ../web/dist/components/*.js ./dist/esm/components && make-dir ./dist/esm && make-dir ./dist/types",
+    "prebuild": "del-cli --force ./dist && copyfiles -f ../web/dist/components/*.js ./dist/esm/components && copyfiles -f ../../node_modules/@stencil/core/internal/client/index.js ./dist/esm/lib/stencil && copyfiles -f ../../node_modules/@stencil/core/internal/client/shadow-css.js ./dist/esm/lib/stencil && make-dir ./dist/esm && make-dir ./dist/types",
```
daine, line 46: Let's add this to the README.md to give them a heads up about this prebuild step.

packages/react-ssr/scripts/build.ts
```diff
@@ -70,12 +71,47 @@ const createReactWrapperModules = async ({ entryPoints, distRoot }) => {
...
+  };
+  
+  await replaceInFile(componentOptions);
+
+  // Patching stencil
+  // Will need to check if the from property needs to be updated when we update @stencil/core
+  const stencilOptions = {
+    files: './dist/esm/lib/stencil/index.js',
+    from: 'var win = typeof window !== "undefined" ? window : {};',
+    to: "var  win = typeof window !== 'undefined' ? window : globalThis || {};",
```
daine, line 91: nit: lint fix for consistency
```suggestion
  const stencilOptions = {
    files: "./dist/esm/lib/stencil/index.js",
    from: "var win = typeof window !== 'undefined' ? window : {};",
    to: "var  win = typeof window !== 'undefined' ? window : globalThis || {};",
```

packages/react-ssr/scripts/build.ts
```diff
@@ -70,12 +71,47 @@ const createReactWrapperModules = async ({ entryPoints, distRoot }) => {
...
+  };
+
+  try {
+    const results = await replaceInFile(stencilOptions);
+
+    if (results[0].hasChanged) {
+      createReactWrapperModules({
+        entryPoints: readdirSync(COMPONENTS_DIR)
+          .filter((file) => file.startsWith('gcds') && file.endsWith('.js') && !file.endsWith('2.js'))
```
daine, line 100: What is '2.js' here?

daine (changes requested):
I think this could work! Just a few things I noticed.

packages/react-ssr/package.json
ethanWallace: This is only happens on our end when we build the package, do we need to include that in the readme?

packages/react-ssr/scripts/build.ts
ethanWallace: If you look in the `dist/esm/components` folder, some of our components get generated into to files (ex: `gcds-icon.js` and `gcds-icon2.js`). That extra bit on that statement stops the `createReactWrapperModules` from using the files that end in `2.js` to avoid a duplicate export error.

packages/react-ssr/package.json
daine: You're right ~ realizing that now. We don't need to mention the specific details like the prebuild step but just a brief mention that the package uses a patched version of stenciljs would be helpful for transparency.

packages/react-ssr/scripts/build.ts
daine: Thanks! I'm curious why we have those generated files, but it doesn't seem to cause any issues at the moment. It'd be nice to have this written down as a comment explaining why it looks for the `2.js` but it's already in existing code so we can keep it as it is.

daine (approved):
LGTM!

PR #573, opened 2024-06-26
feat: updates to stepper component (breaking change + new tag attribute)

ethanWallace:
# Summary | Résumé

Update `gcds-stepper` to match new [design](https://www.figma.com/design/mh2maMG2NBtk41k1O1UGHV/GC-Design-System?node-id=970-2581&t=RDuu7mOQD7Wmkt5T-0) changing stepper from a caption to a heading with caption.

## Changes

- New tag attribute lets you set heading level h1 | h2 | h3.
- Previously, the heading level was set to h6, this functionality has changed.
- New default slot to pass heading text to stepper

### Old implementation

```html
<gcds-stepper current-step="1" total-steps="4">
</gcds-stepper>
```

### New implementation

```html
<gcds-stepper current-step="1" total-steps="4" tag="h2">
  Section title
</gcds-stepper>
```

Requires tokens from https://github.com/cds-snc/gcds-tokens/pull/278.


packages/web/src/components/gcds-stepper/gcds-stepper.tsx
```diff
@@ -49,17 +54,21 @@ export class GcdsStepper {
...
-          tag="h6"
+          tag={tag}
           class="gcds-stepper"
           margin-top="0"
           margin-bottom="300"
         >
-          {`${i18n[lang].step} ${currentStep} ${i18n[lang].of} ${totalSteps}`}
+          <span class="gcds-stepper__steps">
+            {`${i18n[lang].step} ${currentStep} ${i18n[lang].of} ${totalSteps}`}
+            <gcds-sr-only> : </gcds-sr-only>
```
melaniebmn, line 69: Could you add a comment here with a brief explanation of the a11y benefit, please? :)

daine:
When we merge this commit, could we name it as close as possible to the changes it has on current users? I'm thinking something like:
`feat: updates to stepper component (new tag attribute)`

I love how you've detailed the changes in the PR description! Would it be possible to refine it a bit more? This is the page users get to dive into more info on the changes as it's linked on the CHANGELOG and release description.
> - **_New_** `tag` attribute lets you set heading level `h1 | h2 | h3`.
> - Previously, the heading level was set to `h6`, this functionality has changed.
> - New default slot to pass heading text to stepper

ethanWallace:
Just waiting on the translation for our will not render error.

melaniebmn (commented):
Looks good, will approve when translation comes back

packages/web/src/components/gcds-stepper/gcds-stepper.tsx
```diff
@@ -17,12 +17,33 @@ export class GcdsStepper {
...
    * Defines the current step.
    */
-  @Prop() currentStep!: number;
+  @Prop({ mutable: true }) currentStep!: number;
+  @Watch('currentStep')
+  validateCurrentStep() {
+    if (this.currentStep == undefined) {
+      this.currentStep = 1;
+    } else if (this.currentStep > this.totalSteps) {
+      this.currentStep = this.totalSteps;
```
melaniebmn, line 26: Just tested this and seems like this gets overwritten by the `totalSteps` validation. For example if `totalSteps` is set to `4` and I try to set `currentStep` to 28, it keeps `currentStep` as 28 and also updates `totalSteps` to 28.

ethanWallace: This is a tricky thing to figure out since they both have to validate independently. I guess we need to decide which attribute we want to change in a scenario where the current step is higher than the total steps.

Do we want `currentSteps` to be lowered down to the `totalSteps` value or have `totalSteps` raised up to the `currentStep` value?

melaniebmn: I'd vote lowering `currentStep` for `totalSteps` but I have no reasoning for it so I'm very much good with either :D

daine: Good catch Mel! If `currentStep` is out of bounds I would prefer to error out instead of trying to fix the user's (developer's) issue. I would also keep the range between 1 and x, we'll never have negative steps. So psuedocode something like
```javascript
if(this.currentStep === undefined){
  this.currentStep = 1;
} else if (this.currentStep < 1 || this.currentStep > this.totalSteps) { 
  console.error("Invalid number for curentStep"); or throw an error
}
```

Don't want to cause additional bugs for the user by setting the value to something else (apart from auto-increments?)

Commenting this as I haven't tested the entire code yet -- will do so soon!

packages/web/src/components/gcds-stepper/gcds-stepper.tsx
```diff
@@ -17,12 +17,33 @@ export class GcdsStepper {
...
    */
-  @Prop() totalSteps!: number;
+  @Prop({ mutable: true }) totalSteps!: number;
+  @Watch('totalSteps')
+  validateTotalSteps() {
+    if (this.totalSteps == undefined) {
+      this.totalSteps = 3;
+    } else if (this.totalSteps < this.currentStep) {
+      this.totalSteps = this.currentStep;
+    }
```
daine, line 40: Same comment here as above on validation; I feel like we shouldn't be setting arbitrary values for these because they are a required field, is there a really good reason that we have to set it instead of reporting it as an invalid parameter?

packages/web/src/components/gcds-stepper/gcds-stepper.tsx
```diff
@@ -17,12 +17,33 @@ export class GcdsStepper {
   /**
    * Defines the current step.
    */
-  @Prop() currentStep!: number;
+  @Prop({ mutable: true }) currentStep!: number;
+  @Watch('currentStep')
+  validateCurrentStep() {
+    if (this.currentStep == undefined) {
+      this.currentStep = 1;
```
daine, line 24: Also, I think we shouldn't be setting this here, if we want a default value, we should set it in the declaration instead like:
```
@Prop() currentStep = 1;
@Prop() totalSteps = 3;
```

I think `validateCurrentStep()` and `validateTotalSteps()` should only be scoped to process a boolean (valid: true or false). The conditional checks are:
"Is the currentStep a positive number, and is it lesser than totalSteps?"
"Is totalSteps a positive number and is it bigger or equal to currentStep?"

packages/web/src/components/gcds-stepper/test/gcds-stepper.e2e.ts
```diff
@@ -23,18 +23,30 @@ describe('gcds-stepper a11y tests', () => {
...
       .withRules('color-contrast')
       .analyze();
     const results = await colorContrastTest;
 
+    expect(results.violations.length).toBe(0);
+  });
+  /**
+   * Discernable text in heading
+   */
+  it('heading text', async () => {
```
daine, line 39: Can we add a test for the opposite, with no heading?

packages/web/src/components/gcds-stepper/test/gcds-stepper.spec.tsx
```diff
@@ -22,13 +29,92 @@ describe('gcds-stepper', () => {
...
+        </mock:shadow-root>
+        Section title
+      </gcds-stepper>
+    `);
+  });
+
+  /**
+   * Renders as a heading 3
+   */
+  it('renders steps', async () => {
```
daine, line 101: Is it possible to add a test for no heading here?

daine, line 101: And also add another test to add a heading that's not a simple text (i.e. inside another tag)?

daine, line 101: Can we also add a test when the required fields are not set (i.e. currentStep and totalSteps are not set)? 
One where the currentStep is bigger than totalSteps?
One where either currentStep or totalStep was set to a negative value?
One where either currentStep or totalStep was set to a non-number value?

packages/web/src/components/gcds-stepper/gcds-stepper.tsx
```diff
@@ -45,6 +47,11 @@ export class GcdsStepper {
    */
   @Prop() tag: 'h1' | 'h2' | 'h3' = 'h2';
 
+  /**
+   * State to track if component should render
+   */
+  @State() renderError: boolean = false;
```
daine, line 53: Nice!

melaniebmn (commented):
+1 on the unit tests that @daine mentioned - the rest looks great!

ethanWallace:
I believe I have captured all the ways the stepper would fail to render in the unit tests but let me know if I missed one

daine (commented):
Just a few more requests please then we should be good to go!
- Can we fix the linting? @melaniebmn and I noticed you and I don't seem to be getting the lint complaints, I checked my IDE and it now shows me the lints, could you check yours too?
Screenshot 2024-07-09 at 8 19 47 AM

- When you run the tests or check the overview and structure, 4 of them just shows "renders steps" and I realize we have the description in the comments, could we update those?
Screenshot 2024-07-09 at 8 28 45 AM

Thank you!

melaniebmn (approved):
LGTM

daine (approved):
LGTM!

daine (approved):
LGTM!

melaniebmn (approved):
LGTM

PR #574, opened 2024-06-28
chore(storybook): Fix display of property names in storybook

ethanWallace:
# Summary | Résumé

Noticed a couple properties in storybook were not displaying in the correct format in the signature and pagination. Switch from `propertyName` to `property-name` format.


daine (approved):
LGTM

PR #578, opened 2024-07-03
feat: update gcds-card (breaking change)

ethanWallace:
# Summary | Résumé

Update `gcds-card` to match new design in [Figma](https://www.figma.com/design/mh2maMG2NBtk41k1O1UGHV/GC-Design-System?node-id=5757-12701&t=uCOmrEgTNIpr8j4t-0).

## Changes

- Removed `type` attribute. Card will now only render like the previous `link` type.
- Removed `footer` slot
- `tag` attribute has been changed to `badge`
- Added a new default slot to allow more complex HTML to be used as a card description. For simple text, use the `description` attribute but for more complex formatting, use the new slot.

**Note:** requires tokens from https://github.com/cds-snc/gcds-tokens/pull/283


ethanWallace:
Moved to draft as still finalizing `-tag` attribute

packages/web/src/components/gcds-card/gcds-card.tsx
```diff
@@ -82,51 +97,60 @@ export class GcdsCard {
     this.updateLang();
   }
 
-  private get hasCardFooter() {
-    return !!this.el.querySelector('[slot="footer"]');
+  private get renderDescription() {
```
realdylanzheng, line 100: Great that the card can handle more flexible content, however this concept is not built into the [Card component in the Figma library](https://www.figma.com/design/mh2maMG2NBtk41k1O1UGHV/GC-Design-System?node-id=6322-16439&t=naPMzVAI6TCXr9Ak-4), curious how was this feature was handed off or communicated to be added to code?

Also what happens when the innerHTML and the description are both empty?

ethanWallace: This was a change mentioned within our requirements document made to provide better support based on [this issue](https://github.com/cds-snc/design-gc-conception/issues/927).

If there is no passed content or no description attribute, the card will render without a body of text and just have the card title(headline).

realdylanzheng: Thank you, I was unaware of that issue. The main goal as I understand of the slot is to accommodate rich text, but slot opens up much more possibilities for what can be injected? If so is there a plan to limit it to rich text?

ethanWallace: Currently there is nothing being done to limit what can be passed

realdylanzheng: Is the intent to leave it as a freestyle slot? If so adjustments may be needed on the Figma side

ethanWallace: I originally planned for it to be left freestyle similarly to how HTML lets you do whatever you want and you have to know what's the right thing to do.

But with that being said if we as a team want some validation into what is passed into the slot, I can start looking into that.

realdylanzheng: Makes sense, just seems like team leadership wants to see a level of consistency in our components between dev and design. Also limiting our components to it's designed intent (behaviour, visual look and feel etc) is valuable as well. Anyways thanks Ethan, we can checkin with the team at the next available opportunity.

packages/web/src/components/gcds-card/gcds-card.tsx
```diff
@@ -18,20 +19,15 @@ import i18n from './i18n/i18n';
...
-  @Prop({ reflect: true }) type: 'link' | 'action' = 'link';
-
   /**
    * The card title attribute specifies the title that appears on the card
    */
   @Prop({ reflect: true }) cardTitle!: string;
 
   /**
-   * The title element attribute specifies HTML element the title renders as
+   * The crad title tag attribute specifies HTML element the title renders as
```
melaniebmn, line 28: Small typo:

```suggestion
   * The card title tag attribute specifies HTML element the title renders as
```

packages/web/src/components/gcds-card/stories/gcds-card.stories.tsx
```diff
@@ -123,97 +138,111 @@ const TemplatePlayground = args =>
...
+  cardTitle: 'Card title link',
+  href: '#',
+  cardTitleTag: 'a',
+  badge: '',
+  description: '',
   imgSrc: '',
   imgAlt: '',
-  footer: '',
+  default:
+    'Description or supporting text relating to the headline.',
```
melaniebmn, line 190: Can we use the `<gcds-text/>` component here to try to get the users to use them as well if they are using the slot?

packages/web/src/components/gcds-card/gcds-card.css
```diff
@@ -18,50 +18,56 @@
...
       margin: var(--gcds-card-margin);
     }
 
+    .gcds-badge {
+      background-color: var(--gcds-card-badge-background-color);
+      position: absolute;
+      top: 0;
+      left: 0;
+      padding: var(--gcds-card-badge-padding);
+      text-wrap: nowrap;
```
melaniebmn, line 42: Not sure how much you talked to design about this - but can we maybe find a solution for the potential text-overflow issue if the user enters long badge text? It looks a bit broken in the current overflow example. We could add an ellipsis if we want to keep it on one line or let it overflow to multiple lines. Or third option: set a character limit to ensure it fits on one line.

Current overflow:
Screenshot 2024-07-15 at 4 27 54 PM

Overflow onto multiple lines (for example):
Screenshot 2024-07-15 at 4 31 07 PM

melaniebmn (changes requested):
Mostly looks good, but wondering about how to handle the potential badge text-overflow.

packages/web/src/components/gcds-card/gcds-card.css
ethanWallace: Thanks for flagging this. I will bring it to dev/design sync

realdylanzheng: Thank you for flagging. The intent of the badge on the card is to provide short and concise information to the user typically in the form of a single word or two words (short phrase). Common examples: "New", "Updated", "In stock", "Update available". Given this context the option of a character count would seem appropriate. (e.g. max 20 characters)

RavinaSamaroo: Maybe rather than an arbitrary number like 20 characters, we can instead have the limit be closer to what a full line of text across would be?

packages/web/src/components/gcds-card/gcds-card.tsx
```diff
@@ -82,51 +97,62 @@ export class GcdsCard {
     this.updateLang();
   }
 
-  private get hasCardFooter() {
-    return !!this.el.querySelector('[slot="footer"]');
+  private get renderDescription() {
+    if (this.el.innerHTML.trim() != "") {
+      return (<slot></slot>);
```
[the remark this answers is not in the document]
ethanWallace: It should be fine using an unnamed default slot for formatted text.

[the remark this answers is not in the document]
realdylanzheng: Do you think it might enhance the self-documenting nature of the component "api" (more clarity for devs and maintainers) as well as provide a level of future-proofing for the component in terms of potential additional slots etc?

packages/web/src/components/gcds-card/gcds-card.css
ethanWallace: Validation for values over 20 characters has been added to the `badge` property

packages/web/src/components/gcds-card/gcds-card.tsx
ethanWallace: As decided in design/dev sync, we will keep the slot open accepting the risks with that choice.

packages/web/src/components/gcds-card/gcds-card.tsx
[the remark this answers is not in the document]
ethanWallace: Yes I could see that it could potentially enhance the self-documenting nature of the component.

With previous components we haven't named the slot if the content passed to it is tied to the main use case of the component, especially if it is the only slot available.

packages/web/src/components/gcds-card/gcds-card.tsx
```diff
@@ -63,6 +67,30 @@ export class GcdsCard {
    */
   @State() lang: string;
 
+  /**
+   * State to track validation on badge
+   */
+  @State() badgeError: boolean = false;
```
daine, line 73: I can foresee more errors we would need to introduce in the future, so I think it'll be nice if we can start with a generalized error object, starting here.

Something like:
```
@State() error: Object = null
```
or undefined, basically anything that is falsy

then add as more errors exist like...
```
error = { 'badge': ['the badge attribute has a char limit'] }
```

etc

melaniebmn: I like that idea! Small addition: can we use `error = { 'gcds-card': ['the badge attribute has a char limit'] }` to make it easier to see which component the error applies to and aligns with the new error message from the stepper?

packages/web/src/components/gcds-card/gcds-card.tsx
```diff
@@ -18,20 +20,15 @@ import i18n from './i18n/i18n';
...
 
-  /**
-   * The type attribute specifies how the card renders as a link
-   */
-  @Prop({ reflect: true }) type: 'link' | 'action' = 'link';
-
   /**
    * The card title attribute specifies the title that appears on the card
    */
   @Prop({ reflect: true }) cardTitle!: string;
```
melaniebmn, line 26: Following the new validation idea: should we stop the component from being rendered if there is no card title since the card title is required?

packages/web/src/components/gcds-card/gcds-card.tsx
```diff
@@ -18,20 +20,15 @@ import i18n from './i18n/i18n';
...
 
   /**
-   * The title element attribute specifies HTML element the title renders as
+   * The card title tag attribute specifies HTML element the title renders as
    */
-  @Prop() titleElement: 'h3' | 'h4' | 'h5' | 'h6' | 'a' = 'a';
+  @Prop() cardTitleTag: 'h3' | 'h4' | 'h5' | 'h6' | 'a' = 'a';
```
melaniebmn, line 31: Can we make this property required as well, please?

packages/web/src/components/gcds-card/gcds-card.tsx
ethanWallace: I did something similar to what Daine suggested. Will talk about it with both of you next week

packages/web/src/components/gcds-card/gcds-card.tsx
```diff
@@ -109,7 +127,13 @@ export class GcdsCard {
 
     this.updateLang();
 
+    this.validateCardTitle();
+    this.validateHref();
     this.validateBadge();
+
+    if (this.error.includes('href') || this.error.includes('cardTitle')) {
+      console.error('gcds-card: Error rendering, please check required properties \'card-title\' or \'href\'');
+    }
```
daine, line 136: Is it possible to define this as a function and call it as such `this.validateRequiredProps();` here?

And then below, call the same function again before rendering?

I think it could be possible, with multiple watches like this: https://stenciljs.com/docs/reactive-data

packages/web/src/components/gcds-card/gcds-card.tsx
```diff
@@ -143,42 +167,44 @@ export class GcdsCard {
...
-          ) : (
-            <gcds-link href={href} class="gcds-card__title" {...taggedAttr}>
-              {cardTitle}
-            </gcds-link>
-          )}
-          <div class="gcds-card__description">{renderDescription}</div>
-        </div>
-      </Host>
-    );
+    if ((!error.includes('href') && !error.includes('cardTitle'))) {
```
daine, line 170: Is this possible if we add a required props validation?
`if(!this.validateRequiredProps()){`

packages/web/src/components/gcds-card/gcds-card.tsx
```diff
@@ -109,7 +127,13 @@ export class GcdsCard {
 
     this.updateLang();
 
+    this.validateCardTitle();
+    this.validateHref();
     this.validateBadge();
+
+    if (this.error.includes('href') || this.error.includes('cardTitle')) {
+      console.error('gcds-card: Error rendering, please check required properties \'card-title\' or \'href\'');
```
daine, line 135: One last thing: What's the difference between showing `console.error` in a required validation check vs doing a `throw new Error()` like they show on stencils docs? https://stenciljs.com/docs/properties#prop-validation

daine (changes requested):

packages/web/src/components/gcds-card/gcds-card.tsx
ethanWallace: In my tests if I use `throw new Error()`, it will cause the page to not load fully and not continue rendering other web components. Using `console.error` just has the message in the console but everything else renders fine.

melaniebmn (commented):
Comment only: looks good to me, aside from the open comments from Daine re `errorArr`, etc. +1 to those, but seems like we are nearly there!

packages/web/src/components/gcds-card/gcds-card.tsx
```diff
@@ -75,80 +122,104 @@ export class GcdsCard {
...
     this.lang = assignLanguage(this.el);
 
     this.updateLang();
+
+    this.validateBadge();
+
+    let valid = this.validateRequiredProps();
+
+    if (!valid) {
+      console.error(`${i18n['en'].renderError} | ${i18n['fr'].renderError} `);
```
daine, line 150: We can take the extra step of removing hard-coded values in the error message (i.e. in this case dropping `\'card-title\' or \'href\''`)
```suggestion
      console.error(`${i18n['en'].renderError} | ${i18n['fr'].renderError} `, this.errors);
```

or display it comma-separated:
```suggestion
      console.error(`${i18n['en'].renderError} | ${i18n['fr'].renderError} `, this.errors.toString());
```

packages/web/src/components/gcds-card/gcds-card.tsx
daine: Okay thanks for checking, yeah we don't want to do that

packages/web/src/components/gcds-card/gcds-card.tsx
ethanWallace: That could be confusing since in this case optional properties could be included in the errors. In this use case, `badge` could also be in the errors state

daine: Oh that's right. I was hoping there was a way to separate it and reduce hard coded values so we don't need to have to write it out in the error message for all components in the future.

I wonder if we should just drop the property names from the error message like @melaniebmn suggested? Or keep it if we have unit tests that catch the exact error message we'd need to change if ever the required properties are mismatched with the messaging.

daine:
> Is there any plan to ensure a **backward compatibility** with the version 1 of the card component in order to allow implemeters to plan and execute a full migration towards the version 2 of the component?
> 
> Thanks

@duboisp Thanks for reaching out. At the moment, these updates will cause breaking changes and our current recommendation will be to stick using the pinned version until you are ready to plan and execute a full migration towards the updated version.
We are currently still in an alpha phase, and are currently using a SemVer release version `0.x.x` which will mean that updates will not be stable at the moment.
Once we reach a full release starting with `1.x.x` we will provide full support for backwards compatibility moving forward.

packages/web/src/utils/utils.ts
```diff
@@ -108,3 +108,23 @@ export const emitEvent = (
 
   return true;
 };
+
+// Log validation error for required properties in components
+export const logError = (
+  name,
+  errorArr,
+  optionalAttrsArrToRemove
+) => {
```
daine, line 117: Can we add typescript definitions for the parameters?
```suggestion
export const logError = (
  name: string,
  errorArr: [],
  optionalAttrsArrToRemove?: []
) => {
```

packages/web/src/utils/utils.ts
```diff
@@ -108,3 +108,23 @@ export const emitEvent = (
 
   return true;
 };
+
+// Log validation error for required properties in components
```
daine, line 112: Just a small suggestion to use clarify the params -- the most important thing I'd like to see is the one for the `name`

```suggestion
/* Log validation error for required properties in components
 * @param name - name of the component i.e. <gcds-*>
 * @param errorArr - array of attributes with errors
 * @param optionalAttrsArrToRemove - array of optional attributes with errors to be removed from this error message
 */
```

packages/web/src/utils/utils.ts
```diff
@@ -108,3 +108,23 @@ export const emitEvent = (
...
+  let errors = [...errorArr];
+
+  // remove any potential optional attributes from errors array
+  for (const optionalAttr of optionalAttrsArrToRemove) {
+    if (errors.includes(optionalAttr)) {
+      errors.splice(errors.indexOf(optionalAttr), 1);
+    }
+  }
+
+  console.error(`${name}: ${engMsg} (${errors}) | ${name}: ${frMsg} (${errors})`);
```
daine, line 129: Can we add a unit test in `utils.spec.ts` to test for:
- `name` is provided
- `name` (and all other params are missing)
- `errors` array is empty
- `errors` array contains the exact list of errors from `optionalAttrsArrToRemove`

daine (changes requested):
This new util will definitely help us!!! Thanks for including it @ethanWallace . I think I just have one more request, to add code coverage to the new util function, but once that's in we should be good to go!

melaniebmn (approved):
LGTM

daine (approved):
LGTM -- It looks great! I read the run for the "Run tests" workflow and it was pretty comprehensive & everything was passing. Thank you so much!

melaniebmn (approved):
LGTM

rgalipeau:
Thanks Ethan... reviewing for GCDS Drupal with our team.

PR #581, opened 2024-07-08
fix: remove post install on react ssr package

daine:
# Summary | Résumé

Removes post install on the package.json

This fixes the issue of the install command failing to complete


ethanWallace (approved):
LGTM

Issue #577, opened 2024-07-02
Unable to click gcds-nav-link nested in a nav group within the top nav component in Safari


ethanWallace:
When using the Safari browser, you are unable to click a link nested in a dropdown when using the top nav component. Clicking the link closes the dropdown and does not navigate to the desired page.

## Example

Unable to click the links in the contact us dropdown in Safari.

``` 
<gcds-top-nav label="topbar" alignment="right" lang="en">
      <gcds-nav-link href="/home" slot="home">Home</gcds-nav-link>
      <gcds-nav-link href="/installation">Installation</gcds-nav-link>
      <gcds-nav-link href="/foundations">Foundations</gcds-nav-link>
      <gcds-nav-link href="/components" current>Components</gcds-nav-link>

        <gcds-nav-group
          menu-label="Contact us submenu"
          open-trigger="Contact us"
        >
          <gcds-nav-link href="/contact">Form</gcds-nav-link>
          <gcds-nav-link href="/github">GitHub</gcds-nav-link>
          <gcds-nav-link href="/slack">Slack</gcds-nav-link>
        </gcds-nav-group>
      </gcds-top-nav>
```

Issue #576, opened 2024-07-02
Links do not render underline in Safari


ethanWallace:
In the Safari browser, elements using the link style do not render the underline reducing the perceivability of those strings of text to be interacted with.

![Screenshot 2024-07-02 at 12 23 03 PM](https://github.com/cds-snc/gcds-components/assets/10466431/da383830-f39b-4374-abaa-90820511b61f)

adorayi:
Requires investigation.

PR #583, opened 2024-07-09
fix: Safari a11y issues with links and nav group focus

ethanWallace:
# Summary | Résumé

FIx a couple issues related to how Safari renders and handles focus.

## Links not rendering with underline https://github.com/cds-snc/gcds-components/issues/576

Safari does not accept the `text-decoration` shorthand so switched to using all properties.

## Unable to click nested links in a top-nav nav-group https://github.com/cds-snc/gcds-components/issues/577

Safari handles focus a little differently from other browsers so the dropdown logic logic would fire before the `nav-link` click event could be called. Added a small timeout to the `focusout` listener function to allow the `gcds-nav-link` click to go through.

### DIFF
--- a/packages/web/src/components/gcds-link/gcds-link.css
+++ b/packages/web/src/components/gcds-link/gcds-link.css
@@ -14,9 +14,9 @@
   :host .gcds-link {
     cursor: pointer;
     color: var(--gcds-link-default);
-    text-decoration: underline currentColor
-      var(--gcds-link-decoration-thickness);
-    text-underline-offset: var(--gcds-link-underline-offset);
+    text-decoration-style: solid;
+    text-decoration-color: currentColor;
+    text-decoration-thickness: var(--gcds-link-decoration-thickness);
     transition: all 0.35s;
   }
 }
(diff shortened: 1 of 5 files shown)


packages/web/src/components/gcds-link/gcds-link.css
```diff
@@ -14,9 +14,9 @@
   :host .gcds-link {
     cursor: pointer;
     color: var(--gcds-link-default);
-    text-decoration: underline currentColor
-      var(--gcds-link-decoration-thickness);
-    text-underline-offset: var(--gcds-link-underline-offset);
+    text-decoration-style: solid;
+    text-decoration-color: currentColor;
+    text-decoration-thickness: var(--gcds-link-decoration-thickness);
```
melaniebmn, line 19: Seems like this line is getting deleted by accident here - can we put this back in, please?
`text-underline-offset: var(--gcds-link-underline-offset);`

melaniebmn (changes requested):
1 small change, but it looks good otherwise.

packages/web/src/components/gcds-link/gcds-link.css
ethanWallace: Good catch!

melaniebmn (approved):
Ship it!

PR #585, opened 2024-07-17
chore: update stenciljs to 4.18.3

daine:
# Summary | Résumé
The react-ssr branch uses 4.18.3 and already has the patch for it -- putting that change back into the main first

**Note:**
Just tried the following to slowly update from 4.11.0 to a higher version, but the next minor version is not compatible with the current codebase (tests are failing)
@stencil/core@4.11.0 ✔
- npm run install - OK!
- npm run build - OK!
- npm run test - OK!

-> should we downgrade to 4.11.0 on react-ssr?

@stencil/core@4.12.6 ✔
- npm run install - OK!
- npm run build - OK!
- npm run test - FAIL! Vue tests


daine:
Closing this as it is currently incompatible with the current codebase and will need more modifications

PR #586, opened 2024-07-18
ci: update build for angular

daine:
# Summary | Résumé
Updating the build scripts to prep for merging the react ssr branch. This change was something we worked on the react ssr branch because once we add the react ssr package to the workspace, lerna doesn't process the packages in the correct order.

This makes it so the web components package installs the angular output target package first as it is a dependency.


ethanWallace (approved):
LGTM

PR #597, opened 2024-07-29
fix: roll back changes to the angular build process

daine:
# Summary | Résumé
This PR fixes the changes applied when PR #586 was merged.


daine:
We need to investigate what is causing the build to fail as the CI didn't catch it before the release went out, and it was passing previously when it was in the React SSR branch.

melaniebmn (approved):
LGTM

PR #607, opened 2024-08-08
feat: Add gcds-date-input component

ethanWallace:
**Unit tests for new component and functions still need to be added**
**Requires tokens from https://github.com/cds-snc/gcds-tokens/pull/301**

# Summary | Résumé

The date input element is a component that comes pre-configured with form elements to gather date information from users.

```html
<gcds-date-input
  name="date" 
  legend="Date input" 
  format="full | compact"
  error-message=""
  required=""
  value=""
  hint=""
  validateOn="blur | submit | other"
  validator=""
  disabled=""
></gcds-date-input>
```

## Required fields

The following are required fields that are needed to render the `gcds-date-input`.

- **name**: The name of the form field
- **legend**: The legend text for the fieldset
- **format**: The format in which to display. Expects either `full` or `compact`

## Value format

The `gcds-date-input` inputs and outputs the dates in the following formats `YYYY-MM-DD` or `YYYY-MM` depending on the chosen `format`. If the `value` passed to `gcds-date-input` is not in the given format, the date input will not be pre-filled on load.

## Changes

Development of the date input required changes to some other components and the utility functions already used within the GC Design System.

### `gcds-fieldset`

- Updated CSS to fix padding issue
- Added check for `data-date` attribute to handle validation differently.
- Fixed rendering of (required) text to match Figma design.

### `gcds-input`

- Added `part="input"` to internal HTML input to allow styling by parent components.
- Updated CSS to allow better formatting
- Updated `max-width` calculation to display properly
- Added `Watch` functions to track `aria-invalid` and `aria-description` attributes to allow inheriting of information around validation from parent component.

### `gcds-select`

- Added `part="select"` to internal HTML select to allow styling by parent components.
- Updated CSS to allow better formatting
- Added `Watch` functions to track `aria-invalid` and `aria-description` attributes to allow inheriting of information around validation from parent component.

### `inheritAttributes` function

- Removed section of function that removed attributes from host element to allow information around validation to pass information to the inner shadow elements (wasn't working properly anyway)


packages/web/src/components/gcds-date-input/gcds-date-input.tsx
```diff
@@ -0,0 +1,544 @@
...
+        <option value="03">{i18n[lang].march}</option>
+        <option value="04">{i18n[lang].april}</option>
+        <option value="05">{i18n[lang].may}</option>
+        <option value="06">{i18n[lang].june}</option>
+        <option value="07">{i18n[lang].july}</option>
+        <option value="08">{i18n[lang].august}</option>
+        <option value="09">{i18n[lang].september}</option>
+        <option value="10">{i18n[lang].october}</option>
+        <option value="11">{i18n[lang].november}</option>
+        <option value="12">{i18n[lang].december}</option>
```
daine, line 476: Could we remove some hard coded values?

```javascript
const options = Array.from({ length: 13 }, (_, i) => (i < 10 ? `0${i}` : `${i}`));

const item = (
  <gcds-select>
    {options.map((option) => (
      <option key={option} value={option}>
        {i18n[lang][months][option]}
      </option>
    ))}
  </select>
);
```

And the `i18n.js` file:
```json
{
  "months": {
    "01": "January",
    "02": "February",
    "03": "March",
    "04": "April",
    "05": "May",
    "06": "June",
    "07": "July",
    "08": "August",
    "09": "September",
    "10": "October",
    "11": "November",
    "12": "December"
  }
}
```

and change the french file accordingly

packages/web/src/components/gcds-date-input/gcds-date-input.tsx
```diff
@@ -0,0 +1,542 @@
...
+        aria-invalid={hasError['day'].toString()}
+        aria-description={hasError['day'] && errorMessage}
+      ></gcds-input>
+    );
+
+    return (
+      <Host name={name} onBlur={() => this.onBlur()}>
+        {this.validateRequiredProps() && (
+          <gcds-fieldset
+            legend={legend}
```
melaniebmn, line 523: I know we had previously agreed to call this prop `legend` because it is a legend, but after all the discussions we had about fieldsets in the last week, I'm wondering if it would make more sense to name it `label` to keep it consistent (less confusing?) with other form fields for our users. What do you think @daine @ethanWallace ?

packages/web/src/components/gcds-fieldset/gcds-fieldset.tsx
```diff
@@ -251,7 +253,11 @@ export class GcdsFieldset {
     this.validateValidator();
 
     // Assign required validator if needed
-    requiredValidator(this.el, 'fieldset');
+    if (this.el.getAttribute('data-date-input')) {
```
melaniebmn, line 256: Could we name the attribute just `date-input` or even just `date`? As far as I know, `date` is not an existing HTML attribute.

packages/web/src/index.html
```diff
@@ -269,6 +269,13 @@
           required
         ></gcds-input>
 
+        <gcds-date-input
+          legend="Date input"
```
melaniebmn, line 273: Sorry, nitpicky but can we add a hint here as well, please? Seems like all the other form fields in the example use a hint.

packages/web/src/components/gcds-select/gcds-select.tsx
```diff
@@ -392,7 +404,14 @@ export class GcdsSelect {
...
             onChange={e => this.handleInput(e, this.gcdsChange)}
-            aria-invalid={hasError ? 'true' : 'false'}
+            aria-invalid={
+              inheritedAttributes['aria-invalid'] === 'true'
+                ? inheritedAttributes['aria-invalid']
+                : errorMessage
+                  ? 'true'
+                  : 'false'
+            }
+            part="input"
```
melaniebmn, line 414: Could we name this part `select` instead of `input` since it's not referring to an actual input?

packages/web/src/components/gcds-fieldset/gcds-fieldset.tsx
daine: I wonder if `data-date` is better? I know `data-*` is for [custom data attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*) so hopefully we can keep that part. Because of that I think we can also drop `*-input`.
Other options:
- `data-value`
- `data-date-value`

packages/web/src/components/gcds-date-input/gcds-date-input.tsx
daine: I'm now wondering if it's best to wait on the final decision for the change to the fieldset component for this. But yes I'm leaning towards the fact that it's probably appropriate to call it a `label` because that's what it is in this case. From what the designers are also proposing, this label also looks different and will have the `medium` style as opposed to `bold`. I will tag you in the slack thread where they mentioned it

packages/web/src/components/gcds-fieldset/gcds-fieldset.tsx
melaniebmn: Oops, I read `date-date`, `data-date` makes a lot more sense haha

packages/web/src/components/gcds-fieldset/gcds-fieldset.css
```diff
@@ -30,6 +34,12 @@
 
       .legend__required {
         margin: var(--gcds-fieldset-legend-required-margin) !important;
+        font: var(--gcds-text-size-body-desktop); /* --gcds-fieldset-legend-required-font-desktop */
```
melaniebmn, line 37: It seems that we are only changing the font-weight here which means we don't have to add desktop and mobile styles. Can we either add a new token for this or use the new fieldset required label token to make this font change instead of the general text token? Keeps it a little bit more semantic.

packages/web/src/components/gcds-fieldset/gcds-fieldset.css
```diff
@@ -30,6 +34,12 @@
...
+        vertical-align: middle;
+
+        @media only screen and (width < 48em) {
+          font: var(--gcds-text-size-body-mobile);
```
melaniebmn, line 41: Same comment here, this change seems redundant if we only need to change the font-weight. Ignore this if we also need to change other font properties to different values than the legend font values.

melaniebmn (changes requested):
Left some comments with suggestions. Thank you for putting so much work into this component!

packages/web/src/components/gcds-date-input/gcds-date-input.tsx
melaniebmn: Just flagging (I don't disagree): If we wait for the final decision on the fieldset component, we will need to hold releasing this component until then. Since the guidance still seems to be requiring some time, it wouldn't be the worst but Abdul did say that one team was waiting on the release of this component.

packages/web/src/components/gcds-fieldset/gcds-fieldset.css
ethanWallace: Sorry I realized I didn't upload all the changes needed to the fieldset CSS for the required label change.

packages/web/src/components/gcds-fieldset/gcds-fieldset.css
ethanWallace: I was originally only going to update font weight but when viewing the design in figma I noticed the (required) label in fieldset uses body font settings. Should be updated to the correct new token now.

packages/web/src/components/gcds-date-input/stories/overview.mdx
```diff
@@ -0,0 +1,66 @@
...
+{/* prettier-ignore */}
+<ul className="resources-list">
+  <li>
+    <gcds-link size="regular" href="https://design-system.alpha.canada.ca/en/components/date-input/" target="_blank">Guidance</gcds-link>
+  </li>
+  <li>
+    <gcds-link size="regular" href="https://github.com/cds-snc/gcds-components/tree/main/packages/web/src/components/gcds-date-input" target="_blank">Github</gcds-link>
+  </li>
+  <li>
+    <gcds-link size="regular" href="https://www.figma.com/file/mh2maMG2NBtk41k1O1UGHV/Canadian-Digital-Service%E2%80%A8---GC-Design-System?type=design&node-id=1097-2733&mode=design&t=EFOg4bxJ2GTiPcMe-0" target="_blank">Figma</gcds-link>
```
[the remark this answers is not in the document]
ethanWallace: Yes sorry, missed updating that link to the proper link. I use the details files as a base when creating new stories 😄

packages/web/src/components/gcds-date-input/gcds-date-input.tsx
```diff
@@ -0,0 +1,542 @@
...
+    const year = (
+      <gcds-input
+        name="year"
+        label={i18n[lang].year}
+        inputId="year"
+        type="number"
+        size={4}
+        disabled={disabled}
+        value={this.yearValue}
+        autocomplete="year"
```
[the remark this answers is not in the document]
ethanWallace: I chose to remove the autocomplete attributes since there is no way to assign them a value that will work across multiple use cases of the date input.

ethanWallace:
Just realized I might have to rework the isValidDate function (again) since the date input will accept values of either `YYYY-MM-DD` and `YYYY-MM` in both formats which it should not do.

melaniebmn (approved):
Ship it!

PR #608, opened 2024-08-13
fix: update small design misalignments in various components

melaniebmn:
# Summary | Résumé

Update small design misalignments in various components:

- Update Storybook error message text to match docs site
- Move integrated content for date-modified, side-nav, and top-nav components into their i18n files
- Update required label to use regular font weight
- Style breadcrumbs as list for smaller screens
- Always left-align button text in details summary


ethanWallace (approved):
LGTM!

PR #612, opened 2024-08-15
fix: change fieldset legend required to span

melaniebmn:
# Summary | Résumé

Teeny tiny PR to update the fieldset legend required from a `strong` tag to a `span` tag since it won't be bold moving forward anymore.


ethanWallace (approved):
Good catch! LGTM

PR #613, opened 2024-08-15
feat: add new main-container prop to container

melaniebmn:
# Summary | Résumé

While working on the basic page template, I noticed that the container component wasn't scaling down the same way as other core layout components like the header and footer.

## Problem

The header and footer components are set to a width of `90%` while all containers in the container component are set to a width of `100%`. They each use a max-width to limit how far they expand, for example, `1140px` for the `xl` sized container. For bigger screens, this didn't show any visual difference to the header and footer components. But for smaller screens, any content within the container goes all the way to the edge of the viewport and doesn't have the 5% margin on the left and right that the header and footer components provide on smaller screens.

## Solution

For most containers, it makes sense to keep the width set to `100%` and only limit the max-width of the container. When the container component is used to wrap the main content of a page though, it is used as one of the core pieces of the page layout and therefore should align with the alignment of the header and footer components. This means it should be set to a width of `90%` to match the header + footer width. I added a new boolean prop called `main-container` which is supposed to only be set to `true` if the container is the main page container.

## Examples

Before:
Screenshot 2024-08-15 at 4 43 07 PM

After:
Screenshot 2024-08-15 at 4 43 21 PM

Note: Let me know if this makes sense and if you have any better suggestions for the prop name! :)


ethanWallace (approved):
This is great! Thank you for adding this!

I am trying to think of a different property name but nothing is coming to me but it LGTM

PR #617, opened 2024-08-20
fix: Margin and spacing issues in gcds-card

ethanWallace:
# Summary | Résumé

Update `gcds-card` CSS and rendering to use new tokens to fix margin and spacing issues.


melaniebmn (approved):
LGTM

PR #624, opened 2024-08-22
fix: Build to allow new tokens to be imported in react-ssr package

ethanWallace:
# Summary | Résumé

Noticed `@cdssnc/gcds-components-react-ssr` was not inheriting all new tokens properly. Fix the build process by including the react-ssr package in the copy CSS script.


melaniebmn (approved):
LGTM

PR #631, opened 2024-09-04
build: Include react-ssr in normal publish workflow

ethanWallace:
# Summary | Résumé

To solve [issue](https://github.com/cds-snc/design-gc-conception/issues/1084) of `@cdssnc/gcds-components-react-ssr` not being tagged as latest on npm, add `react-ssr` to publishing step that publishes all framework packages. This will reduce duplication of code and tag the release as latest.


daine (approved):
LGTM!

PR #642, opened 2024-09-17
fix(gcds-sr-only): Align with WCAG technique C7

duboisp:
# Summary | Résumé

Align the style of `gcds-sr-only` with the [WCAG CSS technique C7](https://www.w3.org/WAI/WCAG22/Techniques/css/C7) and it avoid the 0 pixel technique because it is not recommended according to a [WebAim article about invisible content just for screen reader users](https://webaim.org/techniques/css/invisiblecontent/). 

It also remove the space taken by the component as it must be visually invisible but only visible to screen reader only.

The following tests were automated and integrated to the unit testing of the `<gcds-sr-only>` component

## Component are visually invisible and don't interfere with visual design
1. Put the `<gcds-sr-only>` component in between 2 other visually rendered component.
2. **Expected:** The bottom of the preceding component and the top of the next one should be the same without any space in between.

| Current | With this fix |
| -------- | ------------ |
| You can observe it does takes an undesirable vertical space. | You can observe that vertical space is removed. |
| image | image |

## Ensure the screen reader content are at least one pixel size
1. Select the `<gcds-sr-only>` component in the DOM
2. **Expected:** The element properties `offsetWidth` and `offsetHeight` must be greater than 0.

### DIFF
--- a/packages/web/src/components/gcds-sr-only/gcds-sr-only.css
+++ b/packages/web/src/components/gcds-sr-only/gcds-sr-only.css
@@ -10,10 +10,12 @@
 
 @layer default {
   :host {
-    display: inline-block;
-    width: 0;
-    height: 0;
-    margin: 0;
+    clip-path: inset(100%);
+    clip: rect(1px, 1px, 1px, 1px);
+    height: 1px;
     overflow: hidden;
+    position: absolute;
+    white-space: nowrap;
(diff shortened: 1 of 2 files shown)


ethanWallace (approved):
Looks good to me! Thanks for doing this, this was on my todo list for awhile now.

PR #646, opened 2024-09-18
fix: Fix typos in tokens

ethanWallace:
# Summary | Résumé

The `gcds-nav-link` has a typo when setting the font property (`var(---gcds-nav-link-font)`). It should be `var(--gcds-nav-link-font)`.


amazingphilippe:
I was curious and did a search, there's one more triple dash in css code: https://github.com/search?q=repo%3Acds-snc%2Fgcds-components+---+language%3Acss&type=code

PR #648, opened 2024-09-18
deps: Update @stencil/core to v4.19.2

ethanWallace:
# Summary | Résumé

Update `@stencil/core` to `v4.19.2` from `v4.11.0`.  Chose the `v4.19.2` instead of the latest `v4.21.0` since they released their SSR solution in recent versions and that would change the react output and there are still a few kinks being worked out.

## Changes

Other changes that were required for this update.

- Update patch for unit tests
- Update build script for `@cdssnc/gcds-components-react-ssr` to find correct line to update
- Remove `applyPolyfills` from `plugin.ts` in Vue package as `applyPolyfills` is no longer exported from `@stencil/core`


daine (approved):
LGTM!!!

PR #650, opened 2024-09-25
fix: Update tokens to fix FIP red and date-input legend font weight

ethanWallace:
# Summary | Résumé

Update to latest tokens package to inherit updates for FIP red in signature and legend font weight in date input.


daine (approved):
LGTM!

Issue #643, opened 2024-09-17
bug: Select will only render initial list of options


ethanWallace:
The `gcds-select` component will only render the first set of options passed to it. If there is some logic to change the options within the select, they will not be seen. This differs from the HTML select which will update to the new options list.

[Code example](https://stackblitz.com/edit/vitejs-vite-rldmhu?file=src%2FApp.tsx)

ClementineHahn:
REady to be published today

ethanWallace:
This issue was fixed with the release of version `v0.26.3`.

PR #655, opened 2024-10-02
fix: Update available options in select if changed

ethanWallace:
# Summary | Résumé

Add `mutationObserver` to `gcds-select` to observe if passed `<options>` have been changed. This allows the select component to keep the same behaviour as the HTML select.

[Example of issue using React](https://stackblitz.com/edit/vitejs-vite-rldmhu?file=src%2FApp.tsx)

https://github.com/cds-snc/gcds-components/issues/643


melaniebmn (approved):
LGTM

PR #665, opened 2024-10-09
chore: Add logic to prevent flash on prop change in storybook

ethanWallace:
# Summary | Résumé

Add some logic to add a `min-height` to storybook frame to hopefully prevent large flashes when component re-renders. Seen when viewing components through the iframe view (`/iframe.html?viewMode=docs&demo=true&singleStory=true&id=components-input--events-properties&lang=en`).


melaniebmn (approved):
LGTM

Issue #671, opened 2024-10-21
bug: "Setup Controls" error on code builder


brockhigg10:
### Prerequisites | Prérequis

- [X] I have read the [Contributing Guidelines](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md). | J’ai lu les [Lignes directrices sur les contributions](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md). | J’accepte de me conformer au [Code de conduite](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/cds-snc/gcds-components/issues) that already report this problem, without success. | J’ai vérifié parmi [les tickets (issues) existants](https://github.com/cds-snc/gcds-components/issues) que ce sujet n’avait pas déjà été soumis.

### GC Design System Components Package and Version | Paquet et version des composants de Système de design GC

Website

### Current Behavior | Comportement observé

When you click "setup controls" link on the code builder (see image) ; It produces this error message (see image)
<img width="1056" alt="Screenshot 2024-10-18 at 2 53 04 PM" src="https://github.com/user-attachments/assets/ec39283f-d174-41a4-a904-a3c035e5d38f">
<img width="1193" alt="Screenshot 2024-10-18 at 2 52 56 PM" src="https://github.com/user-attachments/assets/91e11399-25af-4675-8f1d-87aaebbc91fe">

### Expected Behavior | Comportement attendu

I don't know! I was curious to see what happens when you click setup controls.

### Steps to Reproduce | Étapes pour reproduire le bogue

Open code builder on a component. Hover over event so that "setup controls" link appears.  Click "setup controls"

PR #672, opened 2024-10-21
storybook: Remove setup controls link in event table

ethanWallace:
# Summary | Résumé

As mentioned in https://github.com/cds-snc/gcds-components/issues/671, the events table of storybook will render a `Setup controls` link when the table row is hovered. This creates a reusable object to use for all event controls that will prevent the link from rendering.


melaniebmn (approved):
LGTM

Issue #670, opened 2024-10-21
Bug: error message property prevents submission after correction


ethanWallace:
From [Freshdesk ticket 19577](https://cds-snc.freshdesk.com/a/tickets/19577)

> I am trying to get error messages in gcds-input to work with Drupal. When a form is submitted, Drupal checks for validation errors and sets error messages when needed. I have those appearing using the error-message property.
> 
> The problem I am running into is that if the user corrects their input and clicks on the submit button, it will not submit. I presume this is because something in GCDS is preventing submission when there are error messages shown.
> 
> We need a way of bypassing this so that server-side validation can run.
> 
> The documentation mentions "custom validation" but does not say how it would need to be implemented.

**Ping Support once resolved, so we can circle back with client!**

daine:
Still to be released: https://github.com/cds-snc/gcds-components/pull/668

ethanWallace:
Fix released in our [v0.27.0](https://github.com/cds-snc/gcds-components/pull/668) release

PR #673, opened 2024-10-22
fix: error-message property blocking form submission.

ethanWallace:
# Summary | Résumé

As mentioned in https://github.com/cds-snc/gcds-components/issues/670, when a form component has a filled in `error-message` attribute it will prevent the form from being submitted. Allow bypassing this block by using `validate-on="other"`. This should allow teams using server side validation to use the form components normally.


melaniebmn (approved):
LGTM

PR #676, opened 2024-10-25
feat(gcds-file-uploader): Add files prop for ease of access to uploaded files

ethanWallace:
# Summary | Résumé

To match how developers interact with the native HTML file input, add a `files` property to easily access uploaded files to file uploader.

## Example

With HTML file inputs, to access the files uploaded on a change event a developer would just need to use `event.target.files`. With `gcds-file-uploader` it requires using `event.target.shadowRoot.querySelector('input').files` to acheive the same functionality. Adding the `files` prop will allow developers to use the usual method to gain access to the files.

```html
      <gcds-file-uploader
        uploader-id="form-uploader"
        label="Upload file"
        id="file"
        multiple
        required
      ></gcds-file-uploader>

      <script>
        document.getElementById('file').addEventListener('change', (e) => {
          // File-uploader files - new way
          console.log(e.target.files);
          // Input in shadow-dom files - old way
          console.log(e.target.shadowRoot.querySelector('input').files);
        })
      </script>
```

### DIFF
--- a/packages/web/src/components/gcds-file-uploader/readme.md
+++ b/packages/web/src/components/gcds-file-uploader/readme.md
@@ -10,6 +10,7 @@
 | `accept`                  | `accept`        | Defines the file types the file uploader accepts.                           | `string`                                            | `undefined` |
 | `disabled`                | `disabled`      | Specifies if a file uploader element is disabled or not.                    | `boolean`                                           | `false`     |
 | `errorMessage`            | `error-message` | Error message for an invalid file uploader element.                         | `string`                                            | `undefined` |
+| `files`                   | --              | FileList of uploaded files to input                                         | `FileList`                                          | `undefined` |
 | `hint`                    | `hint`          | Hint displayed below the label.                                             | `string`                                            | `undefined` |
 | `label` _(required)_      | `label`         | Form field label.                                                           | `string`                                            | `undefined` |
 | `multiple`                | `multiple`      | Boolean that specifies if the user is allowed to select more than one file. | `boolean`                                           | `undefined` |
--- a/packages/web/src/components/gcds-file-uploader/test/gcds-file-uploader.e2e.ts
+++ b/packages/web/src/components/gcds-file-uploader/test/gcds-file-uploader.e2e.ts
@@ -12,7 +12,7 @@ describe('gcds-file-uploader', () => {
     expect(element).toHaveClass('hydrated');
   });
 
-  it('renders', async () => {
+  it('upload 1 files', async () => {
     const page = await newE2EPage();
     await page.setContent(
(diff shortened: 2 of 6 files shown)


packages/web/src/components/gcds-file-uploader/readme.md
```diff
@@ -10,6 +10,7 @@
 | `accept`                  | `accept`        | Defines the file types the file uploader accepts.                           | `string`                                            | `undefined` |
 | `disabled`                | `disabled`      | Specifies if a file uploader element is disabled or not.                    | `boolean`                                           | `false`     |
 | `errorMessage`            | `error-message` | Error message for an invalid file uploader element.                         | `string`                                            | `undefined` |
+| `files`                   | --              | FileList of uploaded files to input                                         | `FileList`                                          | `undefined` |
```
daine, line 13: Any idea why this and the `value` below are not getting generated properly?

daine: I just read about it a bit more, and while talking with @ethanWallace found out it's because if it's not a primitive type, [stencil doesn't create an attribute for it unless specified](And I misunderstood - it wasn't about the reflect: true at all! But we do use reflect: true here in file-uploader. Its purpose seems to be so your prop & )

daine (changes requested):
This is great, and a much needed feature for a our users!

One thing I'd like us to do is to test it fully, and have tests for it if we can.
A few scenarios I'd like to see:
- ~Developers grabbing the list of files, and doing an append so users can upload files sequentially without clearing the list~ Update: Developers can set the list of files through the property and have it reflected on the UI
- ~Should developers be able to remove certain files, for example if they are not allowing `.exe` files to be uploaded~ Update: this should be handled by the `accept` attribute
- When the user clicks on the Remove file button, the developer should have the correct list of files if the have the `files` property bound to a variable

packages/web/src/components/gcds-file-uploader/test/gcds-file-uploader.e2e.ts
```diff
@@ -11,6 +11,84 @@ describe('gcds-file-uploader', () => {
     const element = await page.find('gcds-file-uploader');
     expect(element).toHaveClass('hydrated');
   });
+
+  it('upload 1 files', async () => {
```
daine, line 15: nit: 
```suggestion
  it('upload 1 file', async () => {
```

daine (commented):
Just one tiny nit comment and could you also run it through the linter? It's showing me linting errors.

Screenshot 2024-10-28 at 4 17 54 PM
TIA!

daine (approved):
LGTM!

Issue #682, opened 2024-11-02
bug: gcds-fieldset seems to break responsive (narrow) layouts with gcds-select


simberman-gc:
### Prerequisites | Prérequis

- [X] I have read the [Contributing Guidelines](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md). | J’ai lu les [Lignes directrices sur les contributions](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md). | J’accepte de me conformer au [Code de conduite](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/cds-snc/gcds-components/issues) that already report this problem, without success. | J’ai vérifié parmi [les tickets (issues) existants](https://github.com/cds-snc/gcds-components/issues) que ce sujet n’avait pas déjà été soumis.

### GC Design System Components Package and Version | Paquet et version des composants de Système de design GC

0.26.3

### Current Behavior | Comportement observé

If you put a gcds-select with very wide content inside a gcds-fieldset and make your viewport very narrow (or set it to "iPhone SE mode" in developer tools), you will see that the entire gcds-fieldset overflows the viewport on the right side. [Demo in a stackblitz here.](https://stackblitz.com/edit/dfo-gcds-vue-testrig-fj6j2u?file=src%2FApp.vue)

Inside the inspector, you can alleviate the problem by adding `width: 100%; text-overflow: clip;` to the "select" inside the shadow root of the gcds-select. This might not be a desirable way to fix the problem however since the issue only appears when using a fieldset :)

### Expected Behavior | Comportement attendu

The contents of the gcds-fieldset should not overflow the right edge of the viewport.

### System Info | Information sur le système

```shell
Vue 3.4. Observed in Edge on Windows.
```

### Steps to Reproduce | Étapes pour reproduire le bogue

[Please review the stackblitz here.](https://stackblitz.com/edit/dfo-gcds-vue-testrig-fj6j2u?file=src%2FApp.vue)

### Code Reproduction URL | URL de reproduction du code

[Demo in a stackblitz here.](https://stackblitz.com/edit/dfo-gcds-vue-testrig-fj6j2u?file=src%2FApp.vue)

### Additional Information | Informations supplémentaires

[...]

PR #683, opened 2024-11-04
fix(gcds-fieldset): Add CSS to allow form components to maintain responsive layout

ethanWallace:
# Summary | Résumé

As mentioned in https://github.com/cds-snc/gcds-components/issues/682, elements in the `gcds-fieldset` would no longer be responsive and overflow with the viewport. The issue just mentions the `gcds-select` but further tests found this was also happening with the other form components (input, textarea, etc). Added `min-inline-size: auto` to the fieldset's CSS as recommended from the issue.


melaniebmn (approved):
LGTM

daine (approved):
LGTM

PR #689, opened 2024-11-14
refactor: only show external icon for link + button when external prop is used

melaniebmn:
# Summary | Résumé

While working on the new page templates for our docs site, I saw an opportunity to improve the behavior of the link and button components, specifically regarding the display of the external icon. Previously, the external icon was shown whenever the `target="_blank"` attribute was used, but it should only appear for links pointing to external, non-canada.ca pages. I updated the code to display the external icon only when the `external` prop is present on either component. This change now allows users to set `target="_blank"` on internal links without triggering the external icon, ensuring more accurate and consistent behavior.


ethanWallace (approved):
LGTM!

packages/web/src/components/gcds-link/test/gcds-link.spec.ts
```diff
@@ -45,10 +45,10 @@ describe('gcds-link', () => {
   it('[french] renders external link with label + icon', async () => {
     const { root } = await newSpecPage({
       components: [GcdsLink],
-      html: `<gcds-link href="https://google.com" target="_blank" lang="fr">External Link</gcds-link>`,
+      html: `<gcds-link href="https://google.com" external lang="fr">External Link</gcds-link>`,
     });
     expect(root).toEqualHtml(`
-      <gcds-link href="https://google.com" target="_blank" lang="fr">
+      <gcds-link href="https://google.com" external lang="fr">
```
daine, line 51: We already have it added to the `index.html` below, but can we also add it to the test? Let's add one text for:
```javascript
// This link opens in a new tab, but is not external.
<gcds-link target="_blank"> 
```

packages/web/src/components/gcds-button/gcds-button.tsx
```diff
@@ -114,6 +114,11 @@ export class GcdsButton {
    */
   @Prop() download: string | undefined;
 
+  /**
+   * Whether the link is external or not
+   */
+  @Prop() external?: boolean = false;
```
daine, line 120: Because this is a prop change, I think it would be a good idea to be specific here since it only applies if the button is `type="link"`


```suggestion
  /**
   * Set to true if the button is meant to link to an external page. Requires property `type` to be set to `"link"` for this to work.
   */
  @Prop() externalLink?: boolean = false;
```

packages/web/src/components/gcds-button/test/gcds-button.spec.ts
```diff
@@ -153,10 +153,10 @@ describe('gcds-button', () => {
   it('renders external link with label + icon', async () => {
     const { root } = await newSpecPage({
       components: [GcdsButton],
-      html: `<gcds-button type="link" button-style="text-only" href="https://google.com" target="_blank">Button Label</gcds-button>`,
+      html: `<gcds-button type="link" href="https://google.com" external>Button Label</gcds-button>`,
     });
     expect(root).toEqualHtml(`
-      <gcds-button type="link" button-style="text-only" href="https://google.com" target="_blank">
+      <gcds-button type="link" href="https://google.com" external>
```
daine, line 159: Can we also add a test for "target=_blank" but are not displaying the external icon?

daine (changes requested):
This is good, just have a few comments! We also may need to loop design in so they can add these properties in the figma components.

packages/web/src/components/gcds-button/gcds-button.tsx
melaniebmn: Good idea!

packages/web/src/components/gcds-button/gcds-button.tsx
```diff
@@ -114,6 +114,12 @@ export class GcdsButton {
    */
   @Prop() download: string | undefined;
 
+  /**
+   * Set to true if the button links to an external page.
+   * Requires the "type" to be set to "link" to work.
+   */
+  @Prop() external?: boolean = false;
```
daine, line 121: Is it possible to separate the new prop from this PR so we don't introduce a change to this component for now? We can add a prop in a separate PR and go through a component update process.

daine (changes requested):
Just one last item that we discussed during sync earlier

PR #692, opened 2024-11-28
fix: Add postbuild script to web package to fix react build

ethanWallace:
# Summary | Résumé

Add a postbuild script that adds `// @ts-expect-error` to two files generated for the react package build. This will allow the react package to build.

**Note**: Targeting the `main` branch for the tests to run


packages/react/scripts/buildPatch.js
```diff
@@ -0,0 +1,43 @@
...
+  {
+    path: path.join(
+      __dirname,
+      '../src/components/stencil-generated/react-component-lib/utils/index.tsx',
+    ),
+    target: 'return React.forwardRef(forwardRef);',
+    replace: '// @ts-expect-error\n  return React.forwardRef(forwardRef);',
+  },
+  {
+    path: path.join(
```
daine, line 14:
```suggestion
    filePath: path.join(
```

packages/react/scripts/buildPatch.js
daine, line 1: Could we also change this to another name? `buildPatch` implies we're building a patch, but we're actually applying patches here within a pre-build script. Will `patchFiles.js` work? Or `patchScript.js`?

packages/react/scripts/buildPatch.js
```diff
@@ -0,0 +1,43 @@
...
+
+    // Write the updated content back to the file
+    fs.writeFile(file.path, result, 'utf8', err => {
+      if (err) {
+        console.error('Error writing to the file:', err);
+        return;
+      }
+    });
+  });
+});
```
daine, line 43: Can we avoid nested callbacks to keep it simpler and avoid callback loops?
Can we also try and make it so it asynchronously just to make it more efficient? We can use either promises or async/await with `.map`?

daine (changes requested):
Initial review - found a few things we can improve a little on, I'll still continue testing

ethanWallace:
Closing this pull request, work will continue in https://github.com/cds-snc/gcds-components/pull/709

PR #721, opened 2025-01-07
fix(gcds-notice): Add accessible labels to identify notice type

ethanWallace:
# Summary | Résumé

To better accommodate screen readers, add text only available to screen readers in the notice headings announcing the notice type.


melaniebmn (approved):
LGTM

PR #722, opened 2025-01-07
ci: pin workflow image to ubuntu 22

ethanWallace:
# Summary | Résumé

Noticed the run-tests workflow was no longer working as it was failing trying to install dependencies for playwright.

## Cause

GitHub has started transitioning the runner `ubuntu-latest` to use Ubuntu v24 instead of Ubuntu v22.04. Playwright has not been updated to work with the latest version of Ubuntu so we need to force using `v22.04` with our `run-tests` workflow.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   build-deploy:
     name: Run Tests
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04.5
     steps:
       - name: Git Checkout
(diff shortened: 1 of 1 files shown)


daine (changes requested):
Thanks for looking into this.

I also found [more info about version 24 here](https://github.com/actions/runner-images/issues/10636), seems there's still a lot of incompatibility issues with it.

Could we also change the PR (commit) to say the specific change? `ci` seems to be the best prefix as well since it's the workflow file. Here's an example:`ci: pin workflow image to ubuntu 22`

daine (approved):

melaniebmn (approved):
LGTM

Issue #720, opened 2025-01-03
bug: Disabled gcds-button Emits Event When Activated via Keyboard Navigation (Enter Key)


Garry-08:
### Prerequisites | Prérequis

- [X] I have read the [Contributing Guidelines](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md). | J’ai lu les [Lignes directrices sur les contributions](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md). | J’accepte de me conformer au [Code de conduite](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/cds-snc/gcds-components/issues) that already report this problem, without success. | J’ai vérifié parmi [les tickets (issues) existants](https://github.com/cds-snc/gcds-components/issues) que ce sujet n’avait pas déjà été soumis.

### GC Design System Components Package and Version | Paquet et version des composants de Système de design GC

cdssnc/gcds-components-vue, version 0.26.2

### Current Behavior | Comportement observé

Expected Accessibility Behavior: A disabled gcds-button is focusable for screen readers, which is correct.
But the Issue is in:
Despite the :disabled="region?.isUsed" condition being true, pressing the EnterKey while focusing the disabled button via keyboard navigation still triggers the @gcdsClick event.

Clicking the same button with a mouse does not emit the event, which is the correct behavior.

### Expected Behavior | Comportement attendu

Event Behavior: A disabled button should not emit events regardless of how it is interacted with (Tab + Enter or mouse click).

### System Info | Information sur le système

```shell
Browser: Chrome, Edge
Operating System: Windows 10
IDE: IntelliJ IDEA
Node.js Version: v20.12.2
npm Version: 10.5.0
```

### Steps to Reproduce | Étapes pour reproduire le bogue

```
<gcds-button
  button-id="delete-button"
  button-role="danger"
  size="small"
  :disabled="region?.isUsed"
  @gcdsClick="openModal"
>
  Delete
</gcds-button>
```

### Additional Information | Informations supplémentaires

Improvement Suggestion: Visually differentiate the focused state of disabled buttons (e.g., different border color or opacity) as currently it is same as the enabled button.

dfo-proto:
[...]

PR #724, opened 2025-01-07
fix(gcds-button): Prevent gcdsClick event if button is disabled

ethanWallace:
# Summary | Résumé

As mentioned in https://github.com/cds-snc/gcds-components/issues/720, the `gcdsClick` event will still fire even after the button has been disabled. Previous work has been made that will prevent HTML form submissions if the button is disabled but preventing the `gcdsClick` event from emitting when disabled will allow easier use in applications.

### DIFF
--- a/.github/workflows/build-components.yml
+++ b/.github/workflows/build-components.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Git Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-      - uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af
+      - uses: actions/setup-node@48b90677b6048efbc723b11a94acb950d3f1ac36
         with:
           node-version: 18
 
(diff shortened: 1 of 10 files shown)


melaniebmn (approved):
LGTM

daine (changes requested):
Is it possible to add a test for the fix?

melaniebmn (approved):
LGTM

daine:
While testing, I'm able to confirm that it does not fire `gcdsClick` anymore, however I think I'm still able to click on a disabled button?

Tested by adding this script on the index.html page in `src/index.html`:
```js
    const button = document.getElementById('myButton');
    button.addEventListener('click', () => {
      console.log("Button was clicked!");
    });
```

similar to how I verify that gcdsClick doesn't get emitted
```js
    const container = document.querySelector('body'); // or any specific container

    container.addEventListener('gcdsClick', (e) => {
      console.log(`Event gcdsClick was emitted by`, e.target, 'with details:', e.detail);
    });
```

Screenshot 2025-01-08 at 11 22 33 PM

🤔 any ideas why?

daine (approved):

Issue #685, opened 2024-11-13
bug: Angular Textarea's value can't be changed by FormGroup


AlexisLeblancDFO:
### Prerequisites | Prérequis

- [X] I have read the [Contributing Guidelines](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md). | J’ai lu les [Lignes directrices sur les contributions](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md). | J’accepte de me conformer au [Code de conduite](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/cds-snc/gcds-components/issues) that already report this problem, without success. | J’ai vérifié parmi [les tickets (issues) existants](https://github.com/cds-snc/gcds-components/issues) que ce sujet n’avait pas déjà été soumis.

### GC Design System Components Package and Version | Paquet et version des composants de Système de design GC

@cdssnc/gcds-components-angular:0.22.2

### Current Behavior | Comportement observé

When clearing values from the FormGroup, the new value of the TextArea.
![image](https://github.com/user-attachments/assets/2bd5f52c-7c13-48fe-8fa3-fa15a02051fc)

When the associated FormGroup is reset to empty values, input is cleared:
![image](https://github.com/user-attachments/assets/1ff678c1-43dc-4972-97c2-5157f5c635f7)

  

### Expected Behavior | Comportement attendu

When the form is cleared the value of the text area is also reset

### System Info | Information sur le système

```shell
Npm version: 9.4.1
Angular: 18.1.1
```

### Steps to Reproduce | Étapes pour reproduire le bogue

Create a form to and try to reset that form:
![image](https://github.com/user-attachments/assets/d28d507e-89f0-49ca-abf4-d0aa14795408)
![image](https://github.com/user-attachments/assets/ea4a6b51-c936-4f94-8ac0-75d0425caa48)
![image](https://github.com/user-attachments/assets/c334a8d9-ffc7-44b6-beb4-f49269aff353)

ethanWallace:
Hey @AlexisLeblancDFO, the issue has been addressed in our [0.31.0 release](https://github.com/cds-snc/gcds-components/blob/main/CHANGELOG.md#0310).

PR #730, opened 2025-01-09
fix(gcds-textarea): Properly set value in shadow-root textarea

ethanWallace:
# Summary | Résumé

As mentioned in https://github.com/cds-snc/gcds-components/issues/685, the `gcds-textarea` value would not be properly reset when using Formgroup. After some investigation it was revealed this wasn't directly related to the Angular implementation of `gcds-textarea` and was just a bug within the web component.

When the component's value changes, the value was not properly being transferred to the textarea in the shadow-root of the component so the shadow-root textarea would keep the previous value while the rest of the component would register the new value.

### Code example

```html
      <gcds-textarea
        textarea-id="form-message"
        label="Message"
        hint="This is a hint."
        character-count="400"
        required
        id="text"
      ></gcds-textarea>

      <button type="button" onclick="resettextarea();">Reset textarea value</button>

      <script>
        function resettextarea() {
          let text = document.getElementById('text')
          text.value = '';
        }
      </script>
```

### Instructions

Run through the instructions once on the `main` branch and then on this branch to see the fix.

1. Add the code example to document.
2. Type text into the `gcds-textarea` component.
3. Click the reset button.
4. On `main` branch, the textarea will still show the typed text in the textarea but the character count will be reset. On this branch, the textarea should no longer show any text in the textarea.

### DIFF
--- a/packages/web/src/components/gcds-textarea/gcds-textarea.tsx
+++ b/packages/web/src/components/gcds-textarea/gcds-textarea.tsx
@@ -125,6 +125,14 @@ export class GcdsTextarea {
   @Prop({ mutable: true }) value?: string;
 
   /**
+   * Set value on internal textarea to allow proper resets
+   */
+  @Watch('value')
+  watchValue(val) {
+    this.shadowElement.value = val;
+  }
+
+  /**
    * Array of validators
    */
   @Prop({ mutable: true }) validator: Array<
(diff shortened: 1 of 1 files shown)


melaniebmn (approved):
LGTM, thank you for the test instructions that was super helpful and good catch! :)

PR #398, opened 2024-01-31
refactor: css rewrite link

melaniebmn:
# Summary | Résumé

CSS rewrite for the link component to add new CSS features like layers, etc.

Part of fix for https://github.com/cds-snc/design-gc-conception/issues/627

PR #731, opened 2025-01-14
fix(gcds-link): Display property to allow normal link wrap behaviour

ethanWallace:
# Summary | Résumé

While publishing new content for our site, I noticed our `gcds-link` component does not wrap around in paragraphs like a normal `<a>` tag. Update the reset style to `display: inline` to allow the `gcds-link` to wrap like a default HTML `a` tag.

## How to test

Using the paragraph below, observe how the `gcds-link` wraps around the page while changing the viewport size.

```
<p>GC Design System is a service that is owned and operated by the Canadian Digital Service (CDS) at <gcds-link href="https://www.canada.ca/en/employment-social-development.html" external>Employment and Social Development Canada (ESDC).</gcds-link> This service enables Canadian federal departments and agencies to build and design trusted, predictable and accessible web and application experiences.</p>
```

### DIFF
--- a/packages/web/src/components/gcds-link/gcds-link.css
+++ b/packages/web/src/components/gcds-link/gcds-link.css
@@ -2,7 +2,7 @@
 
 @layer reset {
   :host {
-    display: inline-block;
+    display: inline;
 
     slot {
       display: initial;


daine:
The only thing I can find is that we have already set it to the previous value in the [CSS rewrite](https://github.com/cds-snc/gcds-components/pull/398), so if there are any reasons it would most likely related to that @melaniebmn 🤔

daine (approved):
LGTM but @melaniebmn to verify (please wait for her approval as well)

~One thing I did notice, is that the font size seems to change a smaller viewports, just for the link component.~ Okay so I guess it's because the UF isn't up to date and is not updating my `<p>` font size!

melaniebmn (approved):
I can't find anything written down either of why we changed it. I guess we can change it back if we find something that breaks from this styling.

PR #746, opened 2025-02-06
fix: Remove gcds-fieldset dependancy from gcds-date-input

ethanWallace:
# Summary | Résumé

https://github.com/cds-snc/design-gc-conception/issues/1383

With upcoming changes for the `gcds-fieldset` component, remove the use of the component inside `gcds-date-input`'s shadow-dom and switch to native HTML fieldset.

Requires https://github.com/cds-snc/gcds-tokens/pull/392


melaniebmn (approved):
LGTM - tested in both languages and with different inputs.

PR #760, opened 2025-02-20
feat: Add gcds-radios component

ethanWallace:
# Summary | Résumé

Add new `gcds-radios` component to replace `gcds-radio-group` component. The new `gcds-radio` component removes the dependancy on having the radio group wrapped in a `gcds-fieldset` component by making the fieldset part of the new component and moving all validation to the new component as well. This component should now function similar to all the other form components in GC Design System.

## New component

```html
<gcds-radios
  name="string"
  legend="string"
  options=’[{
    “id”: “”,
    “label”: “”,
    “value”: “”,
    “checked”: “”,
    “hint”: “”,
  }]’
  required="boolean"
  hint="string"
  error-message="string"
  disabled="booleam"
  value="string"
  validator="Array"
  Validate-on="blur || submit || other"
>
</gcds-radios>
```

## Changes from `gcds-radio-group`

- New required property `legend` to define the legend for the built in fieldset in the `gcds-radios` component.
- New `value` property to allow easy access to the value of the checked radio button.
- New `required`, `hint`, `error-message`, `disabled`, `validator` and `validate-on` properties to match functionality of other form components in the GC Design System.

## Note 

- This pull request will be merged into [this branch](https://github.com/cds-snc/gcds-components/tree/feat/fieldset-radio-checkbox-rework) until we are ready for release.
- This pull request requires tokens from https://github.com/cds-snc/gcds-tokens/pull/408 but will not be able to load them locally until we are ready for release. For testing the new component please use `npm link` to load the new tokens.


packages/web/src/components/gcds-radios/gcds-radios.tsx
```diff
@@ -0,0 +1,529 @@
...
+  }
+
+  /**
+   * Name attribute for an input element.
+   */
+  @Prop({ reflect: true, mutable: false }) name!: string;
+
+  @Watch('name')
+  validateName() {
+    if (
```
melaniebmn, line 114: Could we create a helper function to push and remove error messages from the errors array? The same code is repeated for `validateOptions`, `validateName`, and `validateLegend`.

packages/web/src/components/gcds-radios/gcds-radios.tsx
```diff
@@ -0,0 +1,529 @@
...
+            {errorMessage ? (
+              <div>
+                <gcds-error-message id="radios-error" messageId="radios">
+                  {errorMessage}
+                </gcds-error-message>
+              </div>
+            ) : null}
+
+            {this.optionObject &&
+              this.optionObject.map(radio => {
```
melaniebmn, line 440: Would you see any value in moving the code for each radio into a subcomponent (like in breadcrumbs) to reduce the length of this file?

melaniebmn (changes requested):
Added a few suggestions - would it also be possible to update the example currently being used in the `index.html` demo file to use the new <gcds-radios>?

packages/web/src/components/gcds-radios/gcds-radios.tsx
ethanWallace: I don't think it would be worth it to create a new component like `gcds-breadcrumbs-item` but I could try to make a [functional component](https://stenciljs.com/docs/functional-components) for the radio buttons.

melaniebmn: That works too :) Mainly thought it would be nice to break up the code a bit more to make it easier to read as one.

packages/web/src/components/gcds-radios/stories/gcds-radios.stories.tsx
```diff
@@ -0,0 +1,312 @@
...
+};
+
+export const CheckedValue = Template.bind({});
+CheckedValue.args = {
+  name: 'radio',
+  legend: 'Legend',
+  options: `[
+    { "label": "Label for radio 1", "id": "radio1", "value": "radio1"},
+    { "label": "Label for radio 2", "id": "radio2", "value": "radio2"}
+  ]`,
```
daine, line 253:
```suggestion
  options: `[
    { "label": "Label for radio 1", "id": "radio1", "value": "radio1"},
    { "label": "Label for radio 2", "id": "radio2", "value": "radio2"}
  ]`,
```

This is a random choice but we can also test out if giving an object here (instead of a string) will render the component

daine (commented):
@ethanWallace it's looking great so far! I haven't completed my review yet as there are still a few things I'm hoping to check.

But one thing I've tested is that, when I pull this into our documentation website and pass an object to it, it fails to render. To be more clear, we normally do this
https://github.com/cds-snc/gcds-docs/blob/cb46e7d93c42a321d2519547a4b916072bdb7622/src/en/contact/contact.md?plain=1#L54

and I tried to remove the `stringify` to test it (with updated parameters).

Is it possible to have it so we don't have to stringify it anymore?

ethanWallace:
> But one thing I've tested is that, when I pull this into our documentation website and pass an object to it, it fails to render. To be more clear, we normally do this
https://github.com/cds-snc/gcds-docs/blob/cb46e7d93c42a321d2519547a4b916072bdb7622/src/en/contact/contact.md?plain=1#L54
and I tried to remove the stringify to test it (with updated parameters).
Is it possible to have it so we don't have to stringify it anymore?

@daine The reason you can't just pass an object without it being turned into a stying first in that example is because it renders down to just HTML. HTML attributes can't accept anything more complicated than a string, number and boolean. IF you were using the `gcds-radios` in React, it would accept the object just fine because that's built into the framework but HTML just has that limitation.

I will build an example of the `gcds-radio` into the index.html that shows how you would set the `options` property on the component using JS.

packages/web/src/components/gcds-radios/gcds-radios.tsx
ethanWallace: I ended up creating two new util functions we can use in other components for property validation. The first is `handleErrors`.
```js
/* Log validation error for required properties in components
 * @param errors - array of attributes with errors
 * @param propertyName - name of the property being checked
 * @param property - value of the property being checked
 * @param external - boolean value for an external check on property value
 * @returns modified array of errors
 */
const handleErrors = (
  errors: string[],
  propertyName: string,
  property: string | boolean | object,
  external: boolean = false,
);

Example use :  this.errors = handleErrors(this.errors, 'name', this.name);
```

This function will check if the property is empty or just whitespace and add it to the errors array. It will also remove the property from the errors array if the values are valid. I also included a `external` property in the function so if we need a more advanced check for a property like checking if the object is in the proper format, it can be passed to the function to be also included in the validation.

The second function is `isValid`.

```js
/* Compare errors array to required props array
 * @param errors - array of attributes with errors
 * @param requiredProps - array of required properties to check against
 * @returns boolean if no matching errors
 */
export const isValid = (errors: string[], requiredProps: string[]) => {
  const intersection = errors.filter(x => requiredProps.includes(x));

  return intersection.length > 0 ? false : true;
};

Example use: return isValid(this.errors, ['name', 'legend', 'options']);
```

Instead of having multiple if statements checking if a required property is in the errors array, we can pass an array of the required properties and the function will return a boolean if there is any overlap.

melaniebmn: Love it :)

melaniebmn: @ethanWallace could you create a ticket to use these new util functions in other components as well? Want to make sure we do this because it will be a nice cleanup of some repetitive code :)

packages/web/src/components/gcds-radios/gcds-radios.tsx
```diff
@@ -0,0 +1,467 @@
...
+  @Prop({ reflect: false, mutable: true }) errorMessage: string;
+  @Watch('errorMessage')
+  validateErrorMessage() {
+    if (this.disabled) {
+      this.errorMessage = '';
+    } else if (!this.hasError && this.errorMessage) {
+      this.hasError = true;
+    } else if (this.errorMessage == '') {
+      this.hasError = false;
+    }
```
melaniebmn, line 143:
```suggestion
    } else {
      this.hasError = this.errorMessage ? !this.hasError : false;
    }
```

I think we could summarize both `if else` into the above.

packages/web/src/components/gcds-radios/stories/gcds-radios.stories.tsx
```diff
@@ -0,0 +1,312 @@
...
+  ${args.validateOn != 'blur' ? `validate-on="${args.validateOn}"` : null}
+  ${args.lang != 'en' ? `lang="${args.lang}"` : null}
+>
+</GcdsRadios>
+`.replace(/\s\snull\n/g, '');
+
+const TemplatePlayground = args =>
+  `
+<!-- Web component code (Angular, Vue) -->
+<gcds-radio-group
```
melaniebmn, line 154: Looks like this still needs to be updated with the new component.

melaniebmn (changes requested):
2 small comments, looks good otherwise

packages/web/src/components/gcds-radios/stories/gcds-radios.stories.tsx
```diff
@@ -0,0 +1,312 @@
...
+  ${args.lang != 'en' ? `lang="${args.lang}"` : null}
+>
+</GcdsRadios>
+`.replace(/\s\snull\n/g, '');
+
+const TemplatePlayground = args =>
+  `
+<!-- Web component code (Angular, Vue) -->
+<gcds-radios
+  name="${args.name}"
```
melaniebmn, line 155: Seems like the legend is missing here:

```suggestion
  name="${args.name}"
  legend="${args.legend}"
```

melaniebmn (changes requested):
One more small change

packages/web/src/components/gcds-radios/stories/gcds-radios.stories.tsx
ethanWallace: I added the rest of the properties as well, it was weird it only had name and options.

melaniebmn (approved):
LGTM

packages/web/src/components/gcds-radios/test/gcds-radios.e2e.ts
daine, line 1: Could we have a few more tests:
- test on an options variable (array of objects)
- test on a blank options string
- test on an (invalid) options object (just an object)
- test on an empty array
- test event emitters example: emits gcdsChange event on change

daine (changes requested):
Almost there! Thanks for making the changes to make the bindings work correctly! Found a few more things, mainly some ways to harden areas where things might fail, some documentation suggestions and lastly if we could add more tests (see my comment on the tests file) that would be awesome!

packages/web/src/components.d.ts
```diff
@@ -9,10 +9,12 @@ import { Validator, ValidatorEntry } from "./validators";
 import { SpacingValues } from "./utils/types/spacing";
 import { ContentValues, GridGapValues } from "./components/gcds-grid/gcds-grid";
 import { RadioObject } from "./components/gcds-radio-group/gcds-radio-group";
+import { RadioObject as RadioObject1 } from "./components/gcds-radios/radio";
 export { Validator, ValidatorEntry } from "./validators";
 export { SpacingValues } from "./utils/types/spacing";
 export { ContentValues, GridGapValues } from "./components/gcds-grid/gcds-grid";
 export { RadioObject } from "./components/gcds-radio-group/gcds-radio-group";
+export { RadioObject as RadioObject1 } from "./components/gcds-radios/radio";
```
daine, line 17: I wonder why it's duplicated?

ethanWallace: Both components use a `RadioObject` that is exported. This won't happen after we remove `gcds-radio-group`

daine: I think my eyes were tired and thought I was seeing two `imports`, but they seem to be `import`, and then `export`!

daine (approved):
LGTM!

PR #796, opened 2025-04-02
fix: GC signature cut off on mobile devices

ethanWallace:
# Summary | Résumé

https://github.com/cds-snc/design-gc-conception/issues/1528

To address issue of the Government of Canada signature being cut off on mobile devices, adjust the `viewBox` value of the svgs to better accommodate the signature.


melaniebmn (approved):
LGTM

Issue #783, opened 2025-03-21
bug: drag and drop not supported by gcds-file-uploader


radmorecameron:
### Prerequisites | Prérequis

- [x] I have read the [Contributing Guidelines](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md). | J’ai lu les [Lignes directrices sur les contributions](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md). | J’accepte de me conformer au [Code de conduite](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/cds-snc/gcds-components/issues) that already report this problem, without success. | J’ai vérifié parmi [les tickets (issues) existants](https://github.com/cds-snc/gcds-components/issues) que ce sujet n’avait pas déjà été soumis.

### GC Design System Components Package and Version | Paquet et version des composants de Système de design GC

@cdssnc/gcds-components@0.33.0

### Current Behavior | Comportement observé

If you drag amd drop a file from your computer to the `gcds-file-uploader` component then the browser will open the file instead of the file being set on the input.

### Expected Behavior | Comportement attendu

The input should accept dragging and dropping files like the native html `input type='file'` does.

### System Info | Information sur le système

```shell
Edge, using HTML web components with Blazor
```

### Steps to Reproduce | Étapes pour reproduire le bogue

1) go to https://design-system.alpha.canada.ca/en/components/file-uploader/
2) try to drag and drop a file onto the `Choose file` in the preview, the input doesnt select the drag and dropepd file

1) Go to an html page with a `<input type='file'>`
2) drag and drop a file onto the choose file, notice that the input selects the drag and droppepd file

ethanWallace:
Hi @radmorecameron, a fix for this issue has been released in [v0.34.0](https://github.com/cds-snc/gcds-components/blob/main/CHANGELOG.md#0340). Please let us know if you are still experiencing this issue.

radmorecameron:
Hi @ethanWallace, this seems to be working as expected now, thank you 😄

PR #801, opened 2025-04-07
fix(gcds-file-uploader): ability to drag and drop

ethanWallace:
# Summary | Résumé

Fix the drag and drop functionality for `gcds-file-uploader` to allow proper reading of files dropped on the file input.

https://github.com/cds-snc/gcds-components/issues/783

## Fixes

- Add `drop` and `dragover` to file uploader to catch files being dropped on the file input. If the `accept` attribute is in use, the file uploader will not accept any files not of that type.
- Fix proper cursor display on interactive elements of the file uploader
- Fix a type error when removing an uploaded file.

### DIFF
--- a/packages/web/src/components/gcds-file-uploader/gcds-file-uploader.tsx
+++ b/packages/web/src/components/gcds-file-uploader/gcds-file-uploader.tsx
@@ -255,7 +255,7 @@ export class GcdsFileUploader {
 
       this.shadowElement.files = dt.files;
       this.files = dt.files;
-      this.addFilesToFormData(this.shadowElement.files);
+      this.addFilesToFormData(Array.from(this.shadowElement.files));
     }
 
     this.value = [...filesContainer];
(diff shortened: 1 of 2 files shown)


packages/web/src/components/gcds-file-uploader/gcds-file-uploader.tsx
```diff
@@ -333,6 +333,47 @@ export class GcdsFileUploader {
...
 
+  /*
+   * Handle drop event on file uploader
+   */
+  private handleDrop(e) {
+    e.preventDefault();
+
+    const droppedFiles = e.dataTransfer.files;
+
+    if (droppedFiles.length > 0) {
```
melaniebmn, line 344: Small suggestion: `droppedFiles` could be `null`, let's ensure we check for it in the if statement.

```suggestion
    if (droppedFiles && droppedFiles.length > 0) {
```

packages/web/src/components/gcds-file-uploader/gcds-file-uploader.tsx
```diff
@@ -333,6 +333,47 @@ export class GcdsFileUploader {
...
+        if (this.isFileAccepted(file)) {
+          dt.items.add(file);
+        }
+      }
+
+      if (dt.files.length > 0) {
+        this.shadowElement.files = dt.files;
+        this.files = dt.files;
+      }
+    }
```
melaniebmn, line 356: Can we add an error message to give the user some feedback if something goes wrong?

melaniebmn (changes requested):
Added two small comments - looks great otherwise!

packages/web/src/components/gcds-file-uploader/gcds-file-uploader.tsx
ethanWallace: An error message has been added if the `isFileAccepted` function returns `false`

melaniebmn (dismissed):
LGTM

packages/web/src/components/gcds-file-uploader/gcds-file-uploader.tsx
```diff
@@ -333,6 +333,52 @@ export class GcdsFileUploader {
...
+
+    // Focus file input to bring atention to any errors
+    this.shadowElement.focus();
+  }
+
+  /*
+   * Check if dropped files are accpeted file types
```
daine, line 365:
```suggestion
   * Check if dropped files are accepted file types
```

daine (changes requested):
It looks good, however I have a question about the addition of the forced error check on the `accept` attribute when this is not a native behaviour of the `input type="file"` element. Messaging @ethanWallace directly to discuss

melaniebmn (approved):
LGTM

daine (approved):
LGTM!

PR #813, opened 2025-04-16
fix: icon font face declaration

melaniebmn:
# Summary | Résumé

This is a hotfix for the icon font face declaration to load through CDN links to avoid files not loading correctly.


daine (approved):
Tested it by removing the UF in the example, and loading the gcds.css file manually to initiate the font load. 
Screenshot 2025-04-16 at 12 38 02 PM

LGTM

PR #815, opened 2025-04-16
chore: remove icon font files

melaniebmn:
# Summary | Résumé

Remove icon font files from being included in the package since we are now using the CDN links.


package.json
daine, line 1: I think we can remove the dev dependency on `ncp` as well
```
    "ncp": "^2.0.0",
```

melaniebmn: Good catch

daine (approved):

daine:
Hold off merging; will review this next week

daine (approved):

daine:
LGTM  -- we don't use these files in the package anymore so it should be safe to be removed upon build too

Issue #819, opened 2025-04-23
bug:


aunderh2539:
### Prerequisites | Prérequis

- [x] I have read the [Contributing Guidelines](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md). | J’ai lu les [Lignes directrices sur les contributions](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md). | J’accepte de me conformer au [Code de conduite](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/cds-snc/gcds-components/issues) that already report this problem, without success. | J’ai vérifié parmi [les tickets (issues) existants](https://github.com/cds-snc/gcds-components/issues) que ce sujet n’avait pas déjà été soumis.

### GC Design System Components Package and Version | Paquet et version des composants de Système de design GC

0.31.0

### Current Behavior | Comportement observé

Currently when using the GcdsButton disabled property this doesn't seem to block the onClick listener when it is set to false.

### Expected Behavior | Comportement attendu

When a user clicks on a button that has been disabled, the listener should be disabled as well and not execute.                                                              

Developer has identified a fix by using onGcdsClick instead of onClick and this seems to work.

### System Info | Information sur le système

```shell
Browser: Chrome, Edge, Firefox
Operating System: Windows 10
IDE: Visual Studio Code
npm Version: 10.9.0
```

### Steps to Reproduce | Étapes pour reproduire le bogue

1. Click on gcds button that has been disabled.
2. Event still executes even though button is disabled.

ethanWallace:
Hey @aunderh2539 , this issue has been addressed in our [v0.34.2 release](https://github.com/cds-snc/gcds-components/blob/main/CHANGELOG.md#0342).

PR #829, opened 2025-04-29
fix: Prevent click event on disabled button

ethanWallace:
# Summary | Résumé

Prevent `click` event from firing when button is disabled. Added additional logic to remove disabled attribute if `type="link` or `disabled="false"`.

For #819 

## How to test

Using example below

```
<gcds-button button-role="start" onClick="buttonClicked()">Start</gcds-button>

<script>
  function buttonClicked() {
    console.log('clicked')
  }
</script>
```

1. Copy the code example above into a document
2. Click the start button and observe 'clicked' in the console
3. Focus the element and click the button using the keyboard. Observe 'clicked' in the console.
4. Add the disabled attribute to the button and repeat the steps above. The button should no longer log anything to the console.

### DIFF
--- a/packages/web/src/components/gcds-button/gcds-button.css
+++ b/packages/web/src/components/gcds-button/gcds-button.css
@@ -77,10 +77,14 @@
 }
 
 @layer disabled {
-  :host .gcds-button[aria-disabled='true'] {
-    cursor: not-allowed;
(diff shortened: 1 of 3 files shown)


melaniebmn (approved):
Tested & LGTM

PR #843, opened 2025-05-13
chore: update runner for tests to have manual dispatch

daine:
# Summary | Résumé
For PRs that are not against main, the tests workflow file doesn't run. It works fine, but we're preparing to release the changes to the fieldset, checkboxes, and radios, and we should have a way to make sure the tests run on the branch or branches to make it easier to catch and work on. Adding a manual dispatch can allow us to selectively run it on a branch.


melaniebmn (approved):
LGTM

PR #846, opened 2025-05-20
refactor: form validators

ethanWallace:
# Summary | Résumé

Refactor form validation and validators in the GC Design System components to allow more descriptive error messages. Validators will also take validators written in the old format and convert them into the new format to not produce any breaking changes for teams who have written their own validators.

## Changes

- Validators now only have `validate` which returns an object of `{ valid: boolean, reasons: { en: string, fr: string }, errors?: object }`.
- All form components have been updated to have `validate-on` set to `blur` by default
- Tests have been added for each form component to test default validation, a custom validator and a custom validator in th old format.

## How to test

Using the code below, switch between using the old and new format of validators and validate the `gcds-input`.

```html
<gcds-input
  input-id="form-name"
  id="text-input"
  label="Name"
  hint="Please enter your full name."
  required
></gcds-input>

<script>
  // Custom validator to allow validation min length, max length or value between min and max
  const getLengthValidator = (min, max) => {
    // Create errorMessage object
    let errorMessage = {};
    if (min && max) {
      errorMessage['en'] =
        `You must enter between ${min} and ${max} characters`;
      errorMessage['fr'] =
        `French You must enter between ${min} and ${max} characters`;
    } else if (min) {
      errorMessage['en'] = `You must enter at least ${min} characters`;
      errorMessage['fr'] =
        `French You must enter at least ${min} characters`;
    } else if (max) {
      errorMessage['en'] = `You must enter less than ${max} characters`;
      errorMessage['fr'] =
        `French You must enter less than ${max} characters`;
    }
    return {
      validate: value => {
        value = value || '';
        if (min && max) {
          return min <= value.length && value.length <= max;
        }
        if (min) {
          return min <= value.length;
        }
        if (max) {
          return value.length <= max;
        }
        return true;
      },
      errorMessage,
    };
  };


  // Old format
  // function getLengthValidator(min, max) {
  //   // Create errorMessage object
  //   let errorMessage = {};
  //   if (min && max) {
  //     errorMessage['en'] =
  //       `You must enter between ${min} and ${max} characters`;
  //     errorMessage['fr'] =
  //       `French You must enter between ${min} and ${max} characters`;
  //   } else if (min) {
  //     errorMessage['en'] = `You must enter at least ${min} characters`;
  //     errorMessage['fr'] =
  //       `French You must enter at least ${min} characters`;
  //   } else if (max) {
  //     errorMessage['en'] = `You must enter less than ${max} characters`;
  //     errorMessage['fr'] =
  //       `French You must enter less than ${max} characters`;
  //   }
  //   return {
  //     validate: value => {
  //       value = value || '';
  //       let valid = true;
  //       if (min && max) {
  //         valid = min <= value.length && value.length <= max;
  //       }
  //       if (min) {
  //         valid = min <= value.length;
  //       }
  //       if (max) {
  //         valid = value.length <= max;
  //       }
  //       console.log(value, value.length, valid, min, max);
  //       return {
  //         valid,
  //         reason: errorMessage,
  //       };
  //     },
  //   };
  // }

  // Get the text input
  const textInput = document.getElementById('text-input');

  // Assign the validator
  textInput.validator = [getLengthValidator(1, 2)];
</script>
```

### DIFF
--- a/packages/web/src/components/gcds-date-input/test/gcds-date-input.e2e.ts
+++ b/packages/web/src/components/gcds-date-input/test/gcds-date-input.e2e.ts
@@ -262,6 +262,111 @@ it('Validation - Invalid day', async () => {
   ).toEqual(dateInputErrorMessage.en.invalidday);
 });
 
+it('Validation - Custom validator', async () => {
+  const page = await newE2EPage();
+  await page.setContent(
+    '<gcds-date-input legend="Date input" name="date" format="full" required></gcds-date-input>',
+  );
+
+  await page.evaluate(() => {
+    const dateInput = document.querySelector('gcds-date-input');
+
+    const expectYear = (year: string) => {
+      return {
+        validate: (value: string) => {
+          const dates = value.split('-');
+          return {
+            valid: dates[0] === year,
+            reason: {
+              en: `The entered year must be ${year}.`,
+              fr: `The entered year must be ${year}.`,
+            },
+          };
+        },
+      };
+    };
+
+    (dateInput as HTMLGcdsDateInputElement).validator = [expectYear('1991')];
+  });
+
+  await page.waitForChanges();
+
+  await page.select('gcds-date-input >>> gcds-select >>> select', '03');
+  await page.type('gcds-date-input >>> .gcds-date-input__day >>> input', '03');
+  await page.type(
+    'gcds-date-input >>> .gcds-date-input__year >>> input',
+    '199',
+  );
+
+  const dateInput = await page.find('gcds-date-input');
+
+  dateInput.callMethod('validate');
+  await page.waitForChanges();
+
+  expect(
+    (await page.find('gcds-date-input >>> gcds-error-message')).innerHTML,
+  ).toEqual('The entered year must be 1991.');
+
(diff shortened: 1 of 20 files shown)


melaniebmn (dismissed):
Looks and works great! Really amazing that we also get to keep backwards compatibility for the old format!

packages/web/src/components/gcds-date-input/test/gcds-date-input.e2e.ts
```diff
@@ -262,6 +262,111 @@ it('Validation - Invalid day', async () => {
...
+          return dates[0] === year;
+        },
+        errorMessage: {
+          en: `The entered year must be ${year}.`,
+          fr: `The entered year must be ${year}.`,
+        },
+      };
+    };
+
+    // @ts-expect-error Old format of validator is different then new format. Will still run in JS environments
```
daine, line 340: nit: typo
```suggestion
    // @ts-expect-error Old format of validator is different than new format. Will still run in JS environments
```

daine (approved):
LGTM -- love the backwards compatibility, and this improves how you use validators. I tested in with the vue starter app on our examples repo and it works as expected. Found one small typo so I just added it as a nit comment :D

PR #850, opened 2025-05-21
fix: SVG aria labels

### LINKED ISSUE #833: Signature svg aria label doesn't match title ID
Label

`aria-labelledby="signature-title"`

Title tag
```
<title id="signature-en-title">Government of Canada / Gouvernement du Canada</title>
```

https://github.com/cds-snc/gcds-components/blob/main/packages/web/src/components/gcds-signature/assets/sig-blk-en.svg?short_path=07a246b

daine:
# Summary | Résumé
Fixes #833 where the `aria-labelledby` tag doesn't match the ID

We changed a few things around when we [added the SR component to the signature](https://github.com/cds-snc/gcds-components/commit/146e8b29fbd216d655c93266fed0185b540b1712), and changed the `aria-labelledby` to point to the SR-Only component's ID.

Looking into this a bit more, it seems there's some research saying that the best approach to SVGs and a11y with aria labels is to add a title with an ID, and add the `aria-labelledby` linking to the title's ID. 

**Accessible SVGs: Perfect Patterns For Screen Reader Users**

_Pattern 5:_ `<svg> + role="img" + <title>`
_Pattern 8:_ `<svg> + role="img" + <title> + aria-labelledby="[ID]"`

Here's the link to the article: https://www.smashingmagazine.com/2021/05/accessible-svg-patterns-comparison/#best-in-show-1

Note: we might not need the [ID for the SR-Only](https://github.com/cds-snc/gcds-components/blame/main/packages/web/src/components/gcds-signature/gcds-signature.tsx#L123) component after this change.

https://design-system.alpha.canada.ca/en/components/signature/
Inspect the element, check that the `aria-labelledby` points to the SR-Only component ID vs the SVG's title ID


melaniebmn:
> One of the reason the SVG's label was set to the `id` of the `sr-only` component in the `gcds-signature` was to provide a bilingual label with each GC name tagged in the correct language. I don't know if switching to the label to the `title` in the SVG will provide the same experience.

+1 to this

daine:
@ethanWallace @melaniebmn 
> One of the reason the SVG's label was set to the `id` of the `sr-only` component in the `gcds-signature` was to provide a bilingual label with each GC name tagged in the correct language. I don't know if switching to the label to the `title` in the SVG will provide the same experience.

Question 1 - Why is it like this for the signature, but the [wordmark is not](https://github.com/cds-snc/gcds-components/blob/main/packages/web/src/components/gcds-signature/assets/wmms-spl-en.svg?short_path=96e1995)? Shouldn't they use the same pattern?

Question 2 - I just tested it, it functions exactly the same way using MacOS VoiceOver in Chrome. The title is bilingual for the SVG image which is the same as SR only. Voiceover on SRonly also just reads the entire text "Government of Canada / Gouvernement du Canada" (and vice versa for french). Doing it this way is fine, we can keep it, but it's probably not a good idea to keep the ID in the `title` of the SVG especially since it's not necessary. It creates confusion having a very similar name with the `sr-only` component. Should we remove it?

melaniebmn (changes requested):
Overall looks good, but the tests for the signature component still need to be updated.

PR #859, opened 2025-06-03
fix: align structural markup for link and search with GCWeb

melaniebmn:
# Summary | Résumé

Align the structural markup of the GCDS `links` and `search` components with GCWeb patterns to ensure consistency when users (developers) use the two together on the same page.

## Changes

- `gcds-link`: `role="link"` has been removed from the `<a>` tag.
- `gcds-search`: If the `suggested` property is not set or assigned, the list attribute on the input element is removed.
- `gcds-search`: Replace the `div` with a `section` for a more semantic markup.

## Zenhub ticket

This is the [Zenhub ticket](https://app.zenhub.com/workspaces/gc-design-system-6100624a19f4cf000e46e458/issues/gh/cds-snc/design-gc-conception/1647) for the change.

### DIFF
--- a/packages/web/src/components/gcds-link/gcds-link.tsx
+++ b/packages/web/src/components/gcds-link/gcds-link.tsx
@@ -182,7 +182,6 @@ export class GcdsLink {
       <Host>
         {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events */}
         <a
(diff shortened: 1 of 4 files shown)


ethanWallace (approved):
LGTM

PR #870, opened 2025-06-06
test: Convert E2E tests to use playwright

ethanWallace:
# Summary | Résumé

With the latest update to `@stencil/core: 4.31.0` stencil has been moving away from using puppeteer for E2E tests. We have decided to move over to using playwright for E2E tests.

## Changes

- new `test.spec` and `test.e2e` scripts in the web package
- `test` script run both test sub scripts and triggers a build before running.
- New `&.e2e.html` files for each component to complete the e2e tests

## TODO

Move all e2e tests to  `playwright`, missing components:
- [x] Checkboxes
- [x] Date input
- [x] File uploader
- [x] Input
- [x] Radios
- [x] Select
- [x] Textarea

Other:
- [x] Fix  Github test action
- [x] Fix Vue tests
- [x] Remove `playwright` from the new e2e file names and delete the old e2e test files once all tests have been converted

**Note**: we will remove `playwright` from the new e2e file names and delete the old e2e test files once all tests have been converted


daine (approved):
Tested it locally, I had to reinstall playwright but it was easy enough, just did `npx playwright install`.

Ran the github actions before approving, it failed, but it was a timeout issue. I see that the timeout is already set pretty high at `60000ms` I believe. I ran it a second time and it passed. While the timeout could still be an issue in the future, this looks like it's in a pretty good state.

Issue #875, opened 2025-06-13
bug: `gcds-card` is being hovered and interacted with instead of `gcds-nav-link` when the link covers part of the gcds-card


radmorecameron:
### Prerequisites | Prérequis

- [x] I have read the [Contributing Guidelines](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md). | J’ai lu les [Lignes directrices sur les contributions](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md). | J’accepte de me conformer au [Code de conduite](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/cds-snc/gcds-components/issues) that already report this problem, without success. | J’ai vérifié parmi [les tickets (issues) existants](https://github.com/cds-snc/gcds-components/issues) que ce sujet n’avait pas déjà été soumis.

### GC Design System Components Package and Version | Paquet et version des composants de Système de design GC

gcds-components@0.35.0

### Current Behavior | Comportement observé

When you open a gcds-nav-group and hover over a gcds-nav-link that covers part of a gcds-card then clicking on the link will click on the card instead

![demonstration of the bug happening](https://github.com/user-attachments/assets/67215a7c-d099-4054-833f-dd4b7c1e13e5)

### Expected Behavior | Comportement attendu

Clicking on the gcds-link should click the link instead of the card

### System Info | Information sur le système

```shell
Windows 11, Microsoft Edge, no framework
```

### Steps to Reproduce | Étapes pour reproduire le bogue

1) Open a gcds-nav-group,
2) Hover over a link that covers part of a gcds-card
3) notice that the card is hovered and that clicking on the link would click on the card instead

### Code Reproduction URL | URL de reproduction du code

https://codesandbox.io/p/sandbox/35qp46?file=%2Findex.html%3A52%2C16

### Additional Information | Informations supplémentaires

Does not happen with mobile navmenu

melaniebmn:
Hey @radmorecameron, thank you so much for opening up an issue. It has been addressed in our [v0.36.0 release](https://github.com/cds-snc/gcds-components/blob/main/CHANGELOG.md#0360). Closing the issue for now, but please feel free to reopen if you think it hasn't been fixed.

PR #877, opened 2025-06-17
Increase `z-index` of `gcds-nav-group` to be above the `z-index` for `gcds-card`

radmorecameron:
# Summary

Stop `gcds-card` from blocking interactions with `gcds-nav-links` in a `gcds-nav-group`
Closes: https://github.com/cds-snc/gcds-components/issues/875

1) Use this page as a reference: https://codesandbox.io/p/sandbox/35qp46?file=%2Findex.html
2) update to use built gcds-components
3) open navlink group
4) interact with the 4th nav link in the group
5) gcds-card is no longer clicked


radmorecameron:
Thanks for the detailed response!

Issue #876, opened 2025-06-16
bug: unchecking a checked checkbox won't remove its value from the `gcdsChange` event if it's checked in the options property


radmorecameron:
### Prerequisites | Prérequis

- [x] I have read the [Contributing Guidelines](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md). | J’ai lu les [Lignes directrices sur les contributions](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md). | J’accepte de me conformer au [Code de conduite](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/cds-snc/gcds-components/issues) that already report this problem, without success. | J’ai vérifié parmi [les tickets (issues) existants](https://github.com/cds-snc/gcds-components/issues) que ce sujet n’avait pas déjà été soumis.

### GC Design System Components Package and Version | Paquet et version des composants de Système de design GC

gcds-components@0.35.0

### Current Behavior | Comportement observé

if a gcds-checkboxes option starts off as checked then unchecking it won't properly update the value for a gcdsChange

### Expected Behavior | Comportement attendu

gcdsChange should say that it was unchecked

### System Info | Information sur le système

```shell
Windows 11, Microsoft Edge, no framework
```

### Steps to Reproduce | Étapes pour reproduire le bogue

1) configure a gcds-checkboxes so that options start off as checked
2) uncheck those checked options
3) notice that gcdsChange event says they're still checked even though the UI says they aren't

![unchecking a checkbox that is configured to be checked doesn't emit the event properly and says that the previously checked values are still checked](https://github.com/user-attachments/assets/ec573270-42f7-4ada-b208-9a5a793ec94c)

### Code Reproduction URL | URL de reproduction du code

https://codesandbox.io/p/sandbox/qw3h83

melaniebmn:
Hey @radmorecameron, thank you so much for opening up an issue. It has been addressed in our [v0.36.0 release](https://github.com/cds-snc/gcds-components/blob/main/CHANGELOG.md#0360). Closing the issue for now, but please feel free to reopen if you think it hasn't been fixed.

PR #878, opened 2025-06-17
fix(gcds-checkboxes): Add logic to make sure a prechecked checkbox's value is removed on change

ethanWallace:
# Summary | Résumé

As mentioned in https://github.com/cds-snc/gcds-components/issues/876, when unchecking an option in a `gcds-checkboxes` component that has been prechecked with the use of the `options`, the value of the checkbox would remain active in the `value`. Add extra logic to the remove value logic to ensure the correct values are contained in the `value` array.

## How to test

Using the code below:

```html
      <form>
        <gcds-checkboxes
          legend="Checkbox group"
          name="checkgroup"
          hint="Select all that apply."
          required
          options='[{ "label": "Checkbox 1 label", "id": "check1", "value": "check1", "checked": true}, { "label": "Checkbox 2 label", "id": "check2", "value": "check2", "checked": true}, { "label": "Checkbox 3 label", "id": "check3", "value": "check3", "checked": true}]'
        ></gcds-checkboxes>

        <gcds-button type="submit">Submit</gcds-button>
      </form>
```

1. Using the `main` branch, add the above code to a page.
2. Uncheck two of the checkboxes and submit the form.
3. Notice all the values are still in the query string on form submission.
4. Using this branch, add the above code to a page.
5. Uncheck two of the checkboxes and submit the form.
6. Notice only the checked value is contained in the query string.

### DIFF
--- a/packages/web/src/components/gcds-checkboxes/gcds-checkboxes.tsx
+++ b/packages/web/src/components/gcds-checkboxes/gcds-checkboxes.tsx
@@ -136,7 +136,6 @@ export class GcdsCheckboxes {
   @Prop({ reflect: true, mutable: true }) value: string | Array<string> = [];
   @Watch('value')
   validateValue(newValue) {
-    console.log(this.name, newValue);
     // Convert string to array
     if (!Array.isArray(newValue)) {
       try {


melaniebmn (approved):
Works like a charm! :)

PR #893, opened 2025-07-03
feat(gcds-input): Add native HTML attributes and validation

ethanWallace:
# Summary | Résumé

Add attributes to match `gcds-input` behaviour to the HTML `input` to provide additional validation.

`gcds-input` has also been expanded to work with the HTML [ValidityState](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState) to provide developers with more options when validating `gcds-input`s and new methods to check current validation satte.

## New properties/attributes

| `autofocus`            | `autofocus`     | If true, the input will be focused on component render                                                                                                                  | `boolean`                                                          | `undefined` |
| `form`                 | `form`          | The ID of the form that the input field belongs to.                                                                                                                     | `string`                                                           | `undefined` |
| `max`                  | `max`           | The maximum value that the input field can accept. Only applies to number input type.                                                                                   | `number \| string`                                                 | `undefined` |
| `maxlength`            | `maxlength`     | The maximum number of characters that the input field can accept.                                                                                                       | `number`                                                           | `undefined` |
| `min`                  | `min`           | The minimum value that the input field can accept. Only applies to number input type.                                                                                   | `number \| string`                                                 | `undefined` |
| `minlength`            | `minlength`     | The minimum number of characters that the input field can accept.                                                                                                       | `number`                                                           | `undefined` |
| `pattern`              | `pattern`       | Specifies a regular expression the form control's value should match. See: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern                         | `string`                                                           | `undefined` |
| `readonly`             | `readonly`      | If true, the input field cannot be modified.                                                                                                                            | `boolean`                                                          | `undefined` |
| `step`                 | `step`          | A number that specifies the granularity that the value must adhere to. Valid for number type. See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#step | `"any" \| number`                                                  | `undefined` |
| `validity`             | `validity`      | Read-only property of the input, returns a ValidityState object that represents the validity states this element is in.                                                 | `ValidityState`                                                    | `undefined` |

## New methods

| checkValidity()  | Check the validity of gcds-input   |  boolean |
| getValidationMessage() | Get validationMessage of gcds-input     |  string  |

## How to test

The tests below will verify the new attributes validate the entered value.

### max/min/step

```html
  <gcds-input
    input-id="form-number"
    label="number"
    name="number"
    type="number"
    min="10"
    max="100"
    step="5"
  ></gcds-input>
```

1. Copy the input above.
2. Enter a number below `10` and remove focus from the input.
3. You should now see an error about minimum value.
4. Enter a number above `100` and remove focus from the input.
5. You should now see an error about maximum value.
6. Enter a `87` and remove focus from the input.
7. You should now see an error about wrong step and suggested values of `85` and `90`.
8. Enter `50` and remove focus the input.
9. You should no longer see an error on the input.

### minlength

```html
  <gcds-input
    input-id="form-name"
    label="Names"
    name="name"
    minlength="5"
  ></gcds-input>
```

1. Copy the input above.
2. Enter a value with 4 characters and remove focus from the input.
3. You should now see an error about minlength.
4. Add an additional character to the value.
5. You should no longer see an error on the input.

### maxlength

```html
  <gcds-input
    input-id="form-name"
    label="Names"
    name="name"
    maxlength="5"
    value="too long"
  ></gcds-input>
```

1. Copy the input above.
2. Remove one letter from the input's value using the input and remove focus from the input.
3. You should now see an error about maxlength.
4. Remove another two characters from the input's value and remove focus form the input.
5. You should no longer see an error on the input.

### pattern

```html
  <gcds-input
    input-id="form-name"
    label="Names"
    name="name"
    pattern="[A-Z]+"
  ></gcds-input>
```

1. Copy the input above.
2. Type one word in lower case and remove focus from the input.
3. You should now see an error about not matching the pattern.
4. Type one word in upper case and remove focus from the input.
5. You should no longer see an error on the input.


packages/web/src/components/gcds-input/gcds-input.tsx
```diff
@@ -318,14 +507,33 @@ export class GcdsInput {
...
     this.initialValue = this.value ? this.value : null;
   }
 
+  componentDidLoad() {
+    let lengthValidity;
+    // maxlength/minlength validation on load
+    if (this.value && (this.minlength || this.maxlength)) {
+      if (this.minlength && this.value.length < this.minlength) {
+        lengthValidity = { tooShort: true };
+      } else if (this.maxlength && this.value.length > this.minlength) {
```
melaniebmn, line 522: Is `this.minlength` supposed to be `this.maxlength`? `tooLong: true` looks like we want to check for a max length here.

melaniebmn (changes requested):
Added a couple of comments / suggestions.

melaniebmn (approved):
LGTM

packages/web/src/components/gcds-input/gcds-input.tsx
```diff
@@ -170,15 +235,11 @@ export class GcdsInput {
    */
   @Watch('aria-invalid')
   ariaInvalidWatcher() {
-    this.inheritedAttributes = inheritAttributes(this.el, this.shadowElement, [
-      'placeholder',
-    ]);
+    this.inheritedAttributes = inheritAttributes(this.el, this.shadowElement);
```
daine, line 238: Just noticed that `placeholder` is here, and I tested it on the current package and we support it

daine (changes requested):
It looks good, but I just tested the placeholder after our conversation this morning and it seems like we used to support it before?

packages/web/src/components/gcds-input/gcds-input.tsx
ethanWallace: The `inheritedAttributes` hasn't really been something we have talked about or documented. We included `placeholder` in the inherited attributes to encourage some adoption from other teams but that didn't happen. I think it is okay if we were to remove it's support fully for now.

PR #918, opened 2025-08-01
build: Set addGlobalStyleToComponents to false

ethanWallace:
# Summary | Résumé

Set `addGlobalStyleToComponents` in the `stencil.config.ts` file to `false` to prevent stencil from adding the global styles CSS file into each component shadowRoot's [adoptedStyleSheet](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/adoptedStyleSheets) to prevent each component from loading more CSS then it needs to.

[More about this feature](https://stenciljs.com/docs/config-extras#addglobalstyletocomponents)


melaniebmn (approved):
LGTM

PR #924, opened 2025-08-12
storybook: Update error summary error links example

ethanWallace:
# Summary | Résumé

Update the `error-links` examples in storybook to include `#` at the start of the `href`. The code preview from the docs site was updated to this but the storybook example was missed.


melaniebmn (approved):
LGTM

PR #940, opened 2025-08-26
docs: add components diagram

daine:
# Summary | Résumé
While reviewing component names and properties, it was easier for me to visualize it when presented as a diagram.

- Renamed `copyCSSFile.js` to `postbuild.js`
- Added path cleanup on the components.json file on `postbuild.js`

This PR fulfills the stretch goals of https://github.com/cds-snc/design-gc-conception/issues/1590

# How to test
1. Run `npm install`
2. Run `npm run build`. This builds everything, including the `components.json` that we will need for the diagram generator.
3. Run `npm run docs:diagram` (can be run from root `/` or `packages/web`). At the moment it only generates the diagram when you run it manually. We can either:
  a. Add it to the build or postbuild script to generate the files, or
  b. Add it to  our GH workflow https://github.com/cds-snc/gcds-components/blob/main/.github/workflows/build-components.yml

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,7 +1,7 @@
 {
-  "packages/angular": "0.41.0",
-  "packages/react": "0.41.0",
-  "packages/vue": "0.41.0",
-  "packages/web": "0.41.0",
-  "packages/react-ssr": "0.41.0-canary.0"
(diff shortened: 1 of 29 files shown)


melaniebmn (approved):
LGTM

PR #952, opened 2025-09-12
fix(select): prevent double-escaping of special characters in select options

melaniebmn:
# Summary | Résumé

A user reported an issue with the `select` component where ampersands (&) in options were being rendered incorrectly.

The issue stems from how the component is rendering the `option` content. The component is using `innerHTML` to set the option content, which includes encoded HTML entities (e.g., `&`). When rendered via JSX, these entities are being escaped again, causing them to appear incorrectly (e.g., `&amp;` instead of `&`).

## What’s Changed

Refactored the rendering logic for the `select` slot to use `textContent` unstead of `innerHTML` to prevent double-escaping of entities like `&`.

## Reported issue

The reported user issue can be found [here](https://cds-snc.freshdesk.com/a/tickets/24269).

## How to test

1. Using the `main` branch, add the following code to your test page:

```
<gcds-select
    select-id="select-props"
    label="Label"
    name="select"
    default-value="Select option."
>
    <option value="1">Option 1</option>
    <option value="2">& Option 2</option>
    <option value="3">Option 3</option>
</gcds-select>
```

2. Click on the select and notice that option 2 displays `&amp;` instead of `&`.
3. Switch to this branch and add the same code.
4. Click on the select and notice that option 2 now correctly displays the `&` as `&`.

### DIFF
--- a/packages/web/src/components/gcds-select/gcds-select.tsx
+++ b/packages/web/src/components/gcds-select/gcds-select.tsx
@@ -448,7 +448,7 @@ export class GcdsSelect {
 
                 return (
                   <option value={opt.getAttribute('value')} {...selected}>
-                    {opt.innerHTML}
+                    {opt.textContent}
                   </option>
                 );
               } else if (opt.nodeName === 'OPTGROUP') {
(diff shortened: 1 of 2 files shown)


ethanWallace (dismissed):
LGTM

daine (approved):
LGTM!

Issue #504, opened 2024-04-18
feat: Autocomplete/searchable input field


simberman-gc:
### Prerequisites

- [X] I've read the [Contributing Guidelines](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md).
- [X] I've searched for [existing issues](https://github.com/cds-snc/gcds-components/issues) that already report this problem, without success.

### Describe the feature request.

We request an input field with an autocomplete/search menu.

### Describe the use case.

Here's the current use case: commercial fishers may hail out from one port and hail in to a different port; we must collect both ports in a logbook, and there are >10000 to choose from. We cannot reliably limit the list except to anticipate that fishers hailing out from the Pacific region will not be hailing in to a port on the Atlantic. We would like to have users start typing the name of a port, search that list server-side, and show only potential matches to the user.

### Describe the preferred solution.

We are perfectly happy to handle the server communication by ourselves. We would be content with an input that is associated with a menu whose DOM elements we can supply and will be visible when the field is focused. Additionally we would like the normal set of display attributes (required, disabled, error message, force-a-match-or-accept-not-a-match etc).

### Describe all alternatives.

* We could dump a 380 kB JSON list of ports into >10000 DOM elements inside a gcds-select, ouch.
* We could implement an extremely formal UI with a search interface and a "select" button, perhaps in a modal popup: 
![image](https://github.com/cds-snc/gcds-components/assets/163869541/dd22a39a-972a-4efe-9568-13e1d3da5b38)
(though this is a lot of overhead for a very simple piece of information)
* We could have the user just type something in, let the system check for an exact match on the list and produce an error if it doesn't (and maybe a success if it does) (the Yahtzee solution)
* We could take a string and do the validation in our internal error-management system, but why?

### Provide related code or examples.

![image](https://github.com/cds-snc/gcds-components/assets/163869541/77a99923-ea7f-4e40-a190-073c29762a73)

simberman-gc:
For reference, an autocomplete field implemented by a governmental entity: [Autocomplete](https://designsystem.gov.scot/components/autocomplete)

adorayi:
Thanks for this. We are reviewing our roadmap to include a component like this for fiscal year 25/26.

adriannelee:
Contribution from @thiessenp-cds:
> Combobox
[...]

PR #970, opened 2025-10-03
feat(input): add new prop `suggestions` for datalist support

radmorecameron:
# Summary | Résumé

This PR allows a datalist to be provided for GcdsInput elements. This can be used as an alternative until https://github.com/cds-snc/gcds-components/issues/504 is properly implemented (for some of our cases at least). Ideally, if something custom were to be implemented it wouldn't require any changes for user's of the gcds-components (just a version bump and they'll be using the new custom autofill).

Although not supported by gcds-input at this time (it seems to work but isn't advertised as working), a datalist works great with input type is range 😄 

I tested with this html markup:
```html
    <gcds-input
      input-id="input-default"
      label="What's your favourite fruit?"
      name="example-default"
      hint="Select your favourite fruit!"
      suggestions='[{ "label": "Apple"}, { "label": "Avocado"}, { "label": "Banana"}, {"label": "Blueberry"}, {"label": "Grape"},{"label": "Strawberry"}, {"label": "Raspberry"}]'
    >
    </gcds-input>
```
Datalists are accessible by default and come baked into the browser.

Range input with suggestions from datalist:

```html
    <gcds-input
      input-id="input-default"
      label="What's your favourite number between 1 and 100?"
      name="example-default"
      type="range"
      min="0"
      max="100"
      hint="Select your favourite number!"
      suggestions='[{ "label": "0"}, { "label": "25"}, { "label": "50"}, { "label": "75"}, { "label": "100"}]'
    >
    </gcds-input>
```

# Demo
a gif of interacting with the favourite fruit autocomplete example using both mouse and keyboard

### DIFF
--- a/.github/workflows/export_github_data.yml
+++ b/.github/workflows/export_github_data.yml
@@ -29,7 +29,7 @@ jobs:
           role-session-name: GithubDataExport
           aws-region: ca-central-1
       - name: Export Data
-        uses: cds-snc/github-repository-metadata-exporter@ccde2e2cc2d351bbc6fcd4146d2193d8da2b7a95
+        uses: cds-snc/github-repository-metadata-exporter@531ae86f67b4c0aa1a40229571211ef73109bda2
         with:
           github-app-id: ${{ secrets.SRE_BOT_RO_APP_ID }}
           github-app-installation-id: ${{ secrets.SRE_BOT_RO_INSTALLATION_ID }}
--- a/packages/react-ssr/package-lock.json
+++ b/packages/react-ssr/package-lock.json
@@ -5000,22 +5000,6 @@
         "node": ">= 0.8.0"
       }
     },
-    "node_modules/prettier": {
-      "version": "3.5.3",
-      "resolved": "https://registry.npmjs.org/prettier/-/prettier-3.5.3.tgz",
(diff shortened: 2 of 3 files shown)


packages/web/specs/components.json
daine, line 1: Let's remove this file from the PR from now, it gets generated on build anyway so it'll get overwritten, what's more important are the code changes

packages/web/src/components/gcds-input/stories/gcds-input.stories.tsx
```diff
@@ -190,6 +205,10 @@ export default {
       action: 'change',
       ...eventProp,
     },
+    gcdsDatalistSelected: {
+      action: 'change',
```
radmorecameron: Should the action be something else?

ethanWallace (approved):
LGTM

packages/web/src/components/gcds-input/suggestion-option.tsx
```diff
@@ -0,0 +1,19 @@
+export type SuggestionOptionObject = {
```
daine:
```suggestion
export type SuggestionOption = {
```

packages/web/src/components.d.ts
```diff
@@ -777,6 +779,10 @@ export namespace Components {
           * A number that specifies the granularity that the value must adhere to. Valid for number type. See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#step
          */
         "step"?: number | 'any';
+        /**
+          * Array of suggestions
```
daine:
```suggestion
          * Array of suggestion options. This creates a datalist element with options to represent permissible or recommended options available to choose from
```

PR #977, opened 2025-10-08
tests(angular): angular framework tests for ngModel and routerLink

ethanWallace:
# Summary | Résumé

Add tests to `packages/angular` to verify `ngModel` and `routerLink` utilities from Angular function as they should.

## About the test app

A new test app has been created in `packages/angular/tests/app`. This app contains four pages with three used to test the various angular features.

### ngModel test page

The `/forms` page lays out all the gcds components that interface with `ngModel`. Each component is given it's own section and once the component has been interacted with the value will appear inside a paragraph below the element saying 'You entered: <value>'.

### routerLink test page

The `/nav-one` and `/nav-two` pages lays out all components that have a `routerLink` directive. These two pages are exact copies of each other using the same angular component, the only difference is that they each component will use the other path to test `routerLink`.

## How to test

1. In the root directory of the repo, run `npm run test`.

### DIFF
--- a/packages/angular/tests/app/src/app/components/navigation/navigation.components.ts
+++ b/packages/angular/tests/app/src/app/components/navigation/navigation.components.ts
@@ -1,12 +1,12 @@
 import { Component, OnInit } from '@angular/core';
 import { CommonModule } from '@angular/common';
-import { ActivatedRoute } from '@angular/router';
+import { ActivatedRoute, RouterModule } from '@angular/router';
 import { GcdsComponentsModule } from '@cdssnc/gcds-components-angular';
 
 @Component({
   selector: 'app-navigation',
   standalone: true,
-  imports: [CommonModule, GcdsComponentsModule],
+  imports: [CommonModule, GcdsComponentsModule, RouterModule],
   template: `
     <gcds-breadcrumbs [hideCanadaLink]="true">
       <gcds-breadcrumbs-item href="/home" routerLink="/home">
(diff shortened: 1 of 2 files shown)


melaniebmn (approved):
LGTM

packages/angular/tests/e2e/ngmodel.spec.ts
```diff
@@ -0,0 +1,68 @@
+import { test, expect } from '@playwright/test';
+
+test('Form component ngModel', async ({ page }) => {
+  await page.goto('http://localhost:4200/forms');
+
```
daine: Let's add initial state tests here, usually it's just to verify that the values are indeed blank before the fille / update happens.

Something like...
```
  await expect(page.locator('#input-name')).toHaveText('');
```

etc. Each form component would need their own initial state test.

packages/angular/tests/e2e/ngmodel.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  // Change gcds-checkboxes
+  await page
+    .locator('gcds-checkboxes input[value="check2"]')
+    .check({ force: true });
+  await expect(page.locator('#input-check')).toHaveText('check2');
+  await page
+    .locator('gcds-checkboxes input[value="check1"]')
+    .check({ force: true });
+  await expect(page.locator('#input-check')).toHaveText('check2,check1');
```
daine: Apart from testing for the initial state, could you also add tests when checkboxes are unchecked?

packages/angular/tests/e2e/ngmodel.spec.ts
daine: Since we only have e2e tests here, I don't see any tests where you would programatically set the ngModel's value and then check if the component / html element's value also changes accordingly. I think it's valuable to have those, could we add them in? I think we'll have to add it to the test app and set the values there.

A few examples:
- set initial state programatically / or on load
- when a button gets pressed, pre-fill an input (or other form component) to a specified value programatically (`component.value = '123'`)
open to other ideas too

Would love to test for clearing the fields / reset

PR #985, opened 2025-10-16
tests(vue): Update Vue tests to test v-model with each component

ethanWallace:
# Summary | Résumé

Update Vue test app in `packages/vue/tests/app` to verify `v-model` functions as it should.

## About the test app

A new test app has been created in `packages/vue/tests/app`. This app contains two pages with one used to test form components properly interact with`v-model`.

### v-model test page

The `/forms` page lays out all the gcds components that interface with `v-model`. Each component is given it's own section and once the component has been interacted with the value will appear inside a paragraph below the element saying 'You entered: <value>'.

## How to test

1. In the root directory of the repo, run `npm run test`.

https://github.com/cds-snc/design-gc-conception/issues/1940

### DIFF
--- a/packages/vue/tests/app/src/router/index.js
+++ b/packages/vue/tests/app/src/router/index.js
@@ -2,11 +2,13 @@ import { createRouter, createWebHistory } from 'vue-router';
 import Home from '../views/Home.vue';
 import Forms from '../views/Forms.vue';
 import FormsPinia from '../views/FormsPinia.vue';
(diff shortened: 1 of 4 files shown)


packages/vue/tests/e2e/v-model.spec.ts
daine: What do you think about renaming this file to `forms.spec.ts` and accommodating the error summary test here too?

daine (approved):
LGTM!

Issue #987, opened 2025-10-17
bug: Date Input's "value" property is only updated when the input is valid


leblanccr:
### Prerequisites | Prérequis

- [x] I have read the [Contributing Guidelines](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md). | J’ai lu les [Lignes directrices sur les contributions](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md). | J’accepte de me conformer au [Code de conduite](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/cds-snc/gcds-components/issues) that already report this problem, without success. | J’ai vérifié parmi [les tickets (issues) existants](https://github.com/cds-snc/gcds-components/issues) que ce sujet n’avait pas déjà été soumis.

### GC Design System Components Package and Version | Paquet et version des composants de Système de design GC

gcds-components-vue@0.43.1

### Current Behavior | Comportement observé

The Date Input component's `value` property is always `null` unless the input is valid. This becomes an issue when the field is optional, because the application cannot distinguish between empty input (a valid scenario) and incomplete/invalid input (an invalid scenario).

For example, the `value` property is `null` in both scenarios pictured below, which prevents the app from being able to properly validate this field.

<img width="481" height="170" alt="Image" src="https://github.com/user-attachments/assets/263b3dee-9784-4a15-a633-a989b90fc907" />
<img width="526" height="175" alt="Image" src="https://github.com/user-attachments/assets/b71d865f-1567-4e65-9990-e6703b878c95" />

### Expected Behavior | Comportement attendu

The `value` property should expose the date string at all times, including when the input is incomplete or invalid. That way, applications using this component are able to distinguish between when the input is empty and when it is invalid, allowing for proper validation of the field.

### System Info | Information sur le système

```shell
The bug was discovered using Nuxt 3.15.4 on Chrome.
```

### Steps to Reproduce | Étapes pour reproduire le bogue

Watch the Date Input component's `value` property while entering either invalid input or no input.

PR #989, opened 2025-10-20
fix(gcds-date-input): Assign value at all times

ethanWallace:
# Summary | Résumé

As mentioned in https://github.com/cds-snc/gcds-components/issues/987, the `gcds-date-input` component only assigned it's value property when the entered information was for a valid date. This was leading to issues because applications using this component were not are able to distinguish between when the input is empty and when it is invalid, allowing for proper validation of the field.

This has been changed to assign the value property at all times and only validate if the information is proper when the component is `required`.

## How to test

Using the HTML below:

```html
        <gcds-date-input
          legend="Date input"
          name="date-input"
          format="full"
          hint="Formatted: YYYY-MM-DD"
          value="12345-12-45"
        ></gcds-date-input>
```

1. In the main branch, read the `value` property of the above date-input using JS.
2. Observe the value should return `null`.
3. Switch to this branch, read the `value` property of the above date-input using JS.
4. Observe the value should now be assigned to `12345-12-45` even though it is not a proper date.

### DIFF
--- a/packages/web/src/components/gcds-date-input/test/gcds-date-input.e2e.ts
+++ b/packages/web/src/components/gcds-date-input/test/gcds-date-input.e2e.ts
@@ -53,13 +53,13 @@ test.describe('gcds-date-input', () => {
     await page.waitForChanges();
 
     await element.locator('select').selectOption('03');
-    await element.locator('input[name="year"]').fill('12345');
+    await element.locator('input[name="year"]').fill('1234');
 
     const value = await element.evaluate(
       el => (el as HTMLGcdsDateInputElement).value,
     );
 
-    expect(value).toEqual('12345-03');
+    expect(value).toEqual('1234-03');
   });
   /**
    * Invalid value
@@ -97,13 +97,13 @@ test.describe('gcds-date-input', () => {
     await page.waitForChanges();
 
     await element.locator('select').selectOption('03');
-    await element.locator('input[name="year"]').fill('1234');
+    await element.locator('input[name="year"]').fill('12345');
 
     const value = await element.evaluate(
       el => (el as HTMLGcdsDateInputElement).value,
     );
 
-    expect(value).toEqual('1234-03');
+    expect(value).toEqual('12345-03');
   });
   /**
    * Validation


daine (changes requested):
The fix seems to work, however, the function `validateValue()` doesn't actually reflect what it does now. The description for the function also says "Default value for the date input element" so I think we should rethink that function altogether.

I was also looking forward to the tests being added for validity tests, and having tests for the value as well as validity when invalid dates are inputted. Would be nice to have them in the e2e tests and spec tests too!

ethanWallace:
> The fix seems to work, however, the function `validateValue()` doesn't actually reflect what it does now. The description for the function also says "Default value for the date input element" so I think we should rethink that function altogether.
> 
> I was also looking forward to the tests being added for validity tests, and having tests for the value as well as validity when invalid dates are inputted. Would be nice to have them in the e2e tests and spec tests too!

@daine New tests have been updated and the `validateValue` function has been renamed to `watchValue`.  The description "Default value for the date input element" was for the `value` attribute not the `validateValue` function but has been rewritten to better reflect the `value` attribute.

packages/web/src/components/gcds-date-input/test/gcds-date-input.e2e.ts
```diff
@@ -52,6 +52,50 @@ test.describe('gcds-date-input', () => {
...
+    await element.locator('input[name="year"]').fill('12345');
+    await element.locator('input[name="day"]').fill('2e3', { force: true });
+
+    const value = await element.evaluate(
+      el => (el as HTMLGcdsDateInputElement).value,
+    );
+
+    expect(value).toEqual('12345-03-2e3');
+  });
+  test('Format: compact - invalid value', async ({ page }) => {
```
daine: I think this might be the valid value, and the other one has the invalid value? This one has
```
    expect(value).toEqual('1234-03');
```

packages/web/src/components/gcds-date-input/test/gcds-date-input.e2e.ts
```diff
@@ -52,6 +52,50 @@ test.describe('gcds-date-input', () => {
...
     await page.waitForChanges();
 
+    await element.locator('select').selectOption('03');
+    await element.locator('input[name="year"]').fill('12345');
+
+    const value = await element.evaluate(
+      el => (el as HTMLGcdsDateInputElement).value,
+    );
+
+    expect(value).toEqual('12345-03');
```
daine: This test is under:
```
  test('Format: compact - value', async ({ page }) => {
```

should it be under the invalid value test? 😅

daine (commented):
Looks great! Thanks for making the change. Just found a few things where the tests might have been swapped?

daine (approved):
LGTM!

PR #1002, opened 2025-11-03
adding bold to required text for checkboxes radio buttons and labels 

cccs-SJafri:
# Summary | Résumé
changing the font to bold for required text in the GCDS-labels, GCDS-checkboxes, and GCDS-radio buttons. 

the reason for the change is to keep the required text in MCP (my cyber portal external) consistent with MCP-2 (My cyber portal internal) as it is bolded in mcp-internal however the gcds componets used in mcp external do not have bolded text.

1. in the root directory run the command npm build 
2. then cd into packages/web 
3. then run npm run start
4. open the application 
5. scroll down to 'Form elements (including stepper and error summary)'
6. all required text should be in bold except the date input

### DIFF
--- a/.github/workflows/release-generator.yml
+++ b/.github/workflows/release-generator.yml
@@ -19,7 +19,7 @@ jobs:
           app-id: ${{ secrets.CDS_RELEASE_BOT_APP_ID }}
           private-key: ${{ secrets.CDS_RELEASE_BOT_PRIVATE_KEY }}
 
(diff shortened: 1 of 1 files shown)


ethanWallace:
Hey @cccs-SJafri thanks for opening a pull request.

Can you please fill out the reason for the change request in the summary above so our designers can review the change?

Thank you

PR #1006, opened 2025-11-07
fix: compact date-input validation

ethanWallace:
# Summary | Résumé

Couple changes around date-input's validation.

## Wrong error message

With the latest change to the `gcds-date-input` value attribute, the date input would display `Enter the day` when a compact date input had a year with less than 4 digits. 

## Extra error message

Added `validate-on="other"` to the number inputs within `gcds-date-input`. This prevents the year and date inputs from showing a native HTML error when someone enters a value like `2e`.

## How to test

Use the HTML below:

```html
      <gcds-date-input
        legend="Date input"
        name="date-input"
        format="compact"
        hint="Formatted: YYYY-MM"
        required
      ></gcds-date-input>
```

1. Using main branch, add the HTML above to a document
2. Fill in the month, add `12` to the year input and focus away.
3. Notice the `Enter a day` error message applied to the date input.
4. Change the year input's value to `2e` and focus away.
5. Notice the error message that appears on the year input.
6. Switch to this branch.
7. Fill in the month, add `12` to the year input and focus away.
8. Notice the error has changed to `Year must be 4 digits`.
9. Change the year input's value to `2e`
10. Notice the error no longer appears on the year input.

### DIFF
--- a/.github/workflows/s3-backup.yml
+++ b/.github/workflows/s3-backup.yml
@@ -5,6 +5,7 @@ on:
     - cron: "0 6 * * *"
 
 permissions:
+  id-token: write
   contents: read
 
 jobs:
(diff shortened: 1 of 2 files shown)


daine (approved):
Let's revisit the valdidate() function given the different requirements (prop settings) at a later date, this fix is important to patch the issue so let's get it out there!

PR #1010, opened 2025-11-17
feat(input): add `inputmode` HTML prop

radmorecameron:
# Summary

Inputmode can improve the UX on mobile devices.

For example, if an input is numeric but isn't supposed to be incremental then you shouldn't be using the `type='number'` input and `type='text'` should be used instead (see references below on why type='number' shouldn't always be used for numerical data).

# Test instructions

Set inputmode on gcdsinput element and look at the virtual keyboard on th mobile device.

# References

## Why inputmode should be used
https://www.atomica11y.com/accessible-design/input-number/
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/number#using_number_inputs
https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/

## inputmode
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inputMode


daine (dismissed):
LGTM!

PR #1013, opened 2025-11-18
feat(gcds-select): Add native HTML attributes and validation

ethanWallace:
# Summary | Résumé

Add attributes to match `gcds-select` behaviour to the HTML `select` to provide additional validation.

`gcds-select` has also been expanded to work with the HTML [ValidityState](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState) to provide developers with more options when validating `gcds-select`s and new methods to check current validation state.

## New properties/attributes

| `autofocus`            | `autofocus`     | If true, the select will be focused on component render                                                                                                                  | `boolean`                                                          | `undefined` |
| `form`                 | `form`          | The ID of the form that the select field belongs to.                                                                                                                     | `string`                                                           | `undefined` |

## New methods

| checkValidity()  | Check the validity of gcds-select   |  boolean |
| getValidationMessage() | Get validationMessage of gcds-select     |  string  |

https://github.com/cds-snc/design-gc-conception/issues/2022

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,7 +1,7 @@
 {
-  "packages/angular": "0.43.1",
-  "packages/react": "0.43.1",
-  "packages/vue": "0.43.1",
-  "packages/web": "0.43.1",
-  "packages/react-ssr": "0.43.1-canary.0"
+  "packages/angular": "0.44.0",
+  "packages/react": "0.44.0",
+  "packages/vue": "0.44.0",
+  "packages/web": "0.44.0",
+  "packages/react-ssr": "0.44.0-canary.0"
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 # Changelog
 
+## [0.44.0](https://github.com/cds-snc/gcds-components/compare/gcds-components-v0.43.1...gcds-components-v0.44.0)
+
+Released on: 2025-11-25
+
+### :rocket: New Features
+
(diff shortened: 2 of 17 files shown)


packages/web/src/components/gcds-select/gcds-select.tsx
```diff
@@ -90,14 +92,38 @@ export class GcdsSelect {
...
   /**
    * Value for a select element.
    */
   @Prop({ mutable: true }) value?: string;
 
   @Watch('value')
   watchValue(val) {
-    this.internals.setFormValue(val ? val : null);
+    if (this.checkIfValidValue(val)) {
+      this.internals.setFormValue(val ? val : null);
```
daine: The check for this would have been moved up a level, so at this point we just need to set it
```suggestion
      this.internals.setFormValue(val);
```

packages/web/src/components/gcds-select/gcds-select.tsx
```diff
@@ -90,14 +92,40 @@ export class GcdsSelect {
    */
   @Prop({ reflect: true, mutable: false }) defaultValue?: string;
 
+
+  /**
+   * If true, the select will be focused on component render
+   */
+  @Prop({ reflect: true }) autofocus: boolean;
```
daine: Should this be optional too? And/or have an inferred default on declaration?
```suggestion
  @Prop({ reflect: true }) autofocus?: false;
```

daine (approved):
LGTM!

PR #1016, opened 2025-11-19
feat(gcds-file-uploader): Add native HTML attributes and validation

ethanWallace:
# Summary | Résumé

Add attributes to match `gcds-file-uploader` behaviour to the HTML `select` to provide additional validation.

`gcds-file-uploader` has also been expanded to work with the HTML [ValidityState](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState) to provide developers with more options when validating `gcds-file-uploader`s and new methods to check current validation state.

## New properties/attributes

| `autofocus`            | `autofocus`     | If true, the file-uploader will be focused on component render                                                                                                                  | `boolean`                                                          | `undefined` |
| `form`                 | `form`          | The ID of the form that the file-uploader field belongs to.                                                                                                                     | `string`                                                           | `undefined` |

## New methods

| checkValidity()  | Check the validity of gcds-file-uploader   |  boolean |
| getValidationMessage() | Get validationMessage of gcds-file-uploader     |  string  |

https://github.com/cds-snc/design-gc-conception/issues/2026

### DIFF
--- a/packages/web/src/components/gcds-file-uploader/gcds-file-uploader.tsx
+++ b/packages/web/src/components/gcds-file-uploader/gcds-file-uploader.tsx
@@ -364,6 +364,7 @@ export class GcdsFileUploader {
    * Update gcds-file-uploader's validity using internal input
    */
   private updateValidity() {
+    if (!this.shadowElement) return;
     const validity = this.shadowElement.validity;
 
     let validationMessage = null;


daine (changes requested):
Just two minor requests then I think it should be good!

PR #1019, opened 2025-11-21
feat(gcds-radios): Add native HTML attributes and validation

ethanWallace:
# Summary | Résumé

Add attributes to match `gcds-radios` behaviour to the HTML `select` to provide additional validation.

`gcds-radios` has also been expanded to work with the HTML [ValidityState](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState) to provide developers with more options when validating `gcds-radios`s and new methods to check current validation state.

## New properties/attributes

| `autofocus`            | `autofocus`     | If true, the radios will be focused on component render                                                                                                                  | `boolean`                                                          | `undefined` |
| `form`                 | `form`          | The ID of the form that the radios field belongs to.                                                                                                                     | `string`                                                           | `undefined` |

## New methods

| checkValidity()  | Check the validity of gcds-radios  |  boolean |
| getValidationMessage() | Get validationMessage of gcds-radios    |  string  |

https://github.com/cds-snc/design-gc-conception/issues/2023


packages/web/src/components/gcds-radios/gcds-radios.tsx
```diff
@@ -350,7 +416,6 @@ export class GcdsRadios {
 
     this.validateValidator();
 
-    this.inheritedAttributes = inheritAttributes(this.el, this.shadowElement);
```
daine: Is this not needed anymore? I'm trying to understand why we need to remove this line (same for checkboxes)

ethanWallace: Currently the inheritedAttributes is an experimental feature that isn't documented anywhere. In it's current form it can only handle one element, and with these two components being groups of multiple elements, it would require a refactor. I just don't think it's worth it at this time to look into that.

packages/web/src/components/gcds-radios/gcds-radios.tsx
```diff
@@ -13,14 +13,14 @@ import {
   AttachInternals,
 } from '@stencil/core';
 
-import { RadioObject, isRadioObject } from './radio';
+import { RadioObject, isRadioObject, } from './radio';
```
daine: extra comma?
```suggestion
import { RadioObject, isRadioObject } from './radio';
```

PR #1049, opened 2025-12-10
remove(gcds-footer): remove workmark variant from footer

melaniebmn:
# Summary | Résumé

As part of the component API alignment work leading to stable v1.0.0 release, remove `wordmark-variant` property from the `gcds-footer`.

## Design rationale

Using the `white` variant of the `gcds-signature` component within the `gcds-footer` component creates colour contrast problems. Removing the option to use the `white` variant of the `gcds-signature` component ensures a better built-in accessibility for the `gcds-footer` component.


ethanWallace (approved):
LGTM

PR #1068, opened 2025-12-19
feat(gcds-container): introduce align + layout props to improve centered + main-container props

melaniebmn:
# Summary | Résumé

As part of the component API alignment work for the upcoming stable v1.0.0 release, this PR introduces breaking changes to the `gcds-container` component. These updates improve consistency with other components and provide more flexible layout and alignment options.

### `main-container` → `layout`

We are renaming the `main-container` prop to `layout` and are changing it from a `boolean` to now provide two options - `full` and `page` to control how the container aligns with the page layout. When set to `page`, the container uses a max width of 1140px and switches to 90% width on smaller screens to scale consistently with core page layout components such as the `gcds-header` and the `gcds-footer` components. When set to `full`, the container spans the full width (100%) of its parent.

The `main-container` prop has been renamed to `layout` and changed from a `boolean` to an `enum` with two options:

- `page`
   For `page` layouts, the container uses a max width of `1140px` and switches to `90%` width on smaller screens to scale consistently with core page layout components such as the `gcds-header` and the `gcds-footer` components. 

- `full`
   Allows the container to span `100%` of its parent’s width.

This change provides clearer intent and greater flexibility when defining container layouts.

### `centered` → `align`

The `centered` prop has been renamed to `align` and changed from a `boolean` to an `enum` with the following options:

- start
- center
- end

This update aligns the API with other components and offers more precise control over container alignment, improving overall usability and consistency.

### DIFF
--- a/packages/react/tests/app/src/App.tsx
+++ b/packages/react/tests/app/src/App.tsx
@@ -1,23 +1,21 @@
 import { Outlet } from 'react-router-dom';
 
-import { GcdsHeader, GcdsFooter, GcdsContainer } from '@cdssnc/gcds-components-react';
+import {
+  GcdsHeader,
+  GcdsFooter,
+  GcdsContainer,
+} from '@cdssnc/gcds-components-react';
 
 function App() {
   return (
     <>
       <GcdsHeader></GcdsHeader>
-      <GcdsContainer
-        size='xl'
-        tag='main'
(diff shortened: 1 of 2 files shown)


daine:
Could you update the conventional commit to feat? We are changing the API so it's not a refactor :) @melaniebmn

PR #1074, opened 2026-01-05
fix(gcds-search): Suggested attribute assignment

ethanWallace:
# Summary | Résumé

Previously the `suggested` attribute to assign predefined search terms on the `gcds-search` component was only working if the array was assigned using the Javascript property. The component should now accept the array using the `suggested` attribute.

## How to test

```
<gcds-search suggested='["red", "green", "blue"]'></gcds-search>
```

1. On the main branch use the HTML above.
2. Notice the search component does not provide any suggested search terms in a datalist when focused.
3. On this branch using the HTML above.
4. Notice the options 'red', 'green' and 'blue' will now appear when focusing the search input.

### DIFF
--- a/packages/web/specs/components.json
+++ b/packages/web/specs/components.json
@@ -10086,23 +10086,21 @@
         },
         {
           "name": "suggested",
-          "type": "string[]",
+          "type": "string | string[]",
           "complexType": {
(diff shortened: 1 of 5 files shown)


melaniebmn (approved):
LGTM

PR #1079, opened 2026-01-06
remove(react-ssr): Remove @cdssnc/gcds-components-react-ssr package

ethanWallace:
# Summary | Résumé

As part of the work to reach a stable release of `v1.0.0`, remove the `@cdssnc/gcds-components-react-ssr` package from the code base. 

The react-ssr package was added back in July of 2024 as an experimental canary package. The GC Design System team has not been able to commit the time we wanted to be confident that the package is stable and meets the standards of our other packages so we have chosen to remove the package for our `v1.0.0` release.

The GC Design System team will continue to look into some of the tools provided by Stencil to continue to supply SSR ready compnents.


melaniebmn (approved):
LGTM

Issue #1069, opened 2025-12-28
bug: <gcds-date-input> month dropdown disappears when toggling lang between en and fr


tinaapangging:
### Prerequisites | Prérequis

- [x] I have read the [Contributing Guidelines](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md). | J’ai lu les [Lignes directrices sur les contributions](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md). | J’accepte de me conformer au [Code de conduite](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/cds-snc/gcds-components/issues) that already report this problem, without success. | J’ai vérifié parmi [les tickets (issues) existants](https://github.com/cds-snc/gcds-components/issues) que ce sujet n’avait pas déjà été soumis.

### GC Design System Components Package and Version | Paquet et version des composants de Système de design GC

@cdssnc/gcds-components-vue@0.46.0

### Current Behavior | Comportement observé

When toggling the lang prop on <gcds-date-input> from 'en' to 'fr' (and back) at runtime, the month dropdown stops rendering (disappears), while other parts of the component continue to work. This happens consistently when the locale is switched programmatically

**_please check the video attached._**

https://github.com/user-attachments/assets/3998ffeb-cf8f-4d1c-9a79-79b13ce63d3e

### Expected Behavior | Comportement attendu

The month dropdown should render the localized month names for the current lang (en or fr) immediately after toggling.

### System Info | Information sur le système

```shell

```

### Steps to Reproduce | Étapes pour reproduire le bogue

**_The code below is just an example to reproduce the behaviour,_** 

```
<script setup>
	import { ref } from 'vue';
	import { useI18n } from 'vue-i18n';

	const { locale } = useI18n();
	const currentLocale = ref(locale.value);

	function toggleLocale() {
		currentLocale.value = currentLocale.value === 'en' ? 'fr' : 'en';
		locale.value = currentLocale.value;
	}
</script>

<template>
	<gcds-container size="md" centered padding="600">
		<gcds-button @click="toggleLocale">
			{{ currentLocale === 'en' ? 'En' : 'Fr' }}
		</gcds-button>
		<gcds-date-input
[...]

PR #1081, opened 2026-01-06
fix(gcds-select): Remove resetting value on options rerender

ethanWallace:
# Summary | Résumé

As mentioned in https://github.com/cds-snc/gcds-components/issues/1069, the `gcds-date-input` loses the month value when language has been switched on the component. This was caused by logic built into the `gcds-select` that resets the value when the options have changed. Since the most common use case of options changing inside the select component will most likely be language switches, like in the `gcds-date-input`, we have decided tor emove that bit of logic from the `gcds-select`.

## How to test

```html
      <gcds-date-input
        legend="Date input"
        name="date-input"
        format="full"
        hint="Formatted: YYYY-MM-DD"
        lang="en"
      ></gcds-date-input>
```

1. On the main branch using the HTML above, enter a valid date into the date input.
2. Modify the HTML in dev tools and change the `lang` attribute to `fr`.
3. Notice day and year keep their values but the month has been reset.
4. On this branch, enter a valid date into the date input.
5. Modify the HTML in dev tools and change the `lang` attribute to `fr`.
6.  Notice all values have stay assigned.

### DIFF
--- a/packages/web/src/components/gcds-select/gcds-select.tsx
+++ b/packages/web/src/components/gcds-select/gcds-select.tsx
@@ -413,8 +413,6 @@ export class GcdsSelect {
     };
     const observer = new MutationObserver(() => {
       this.options = Array.from(this.el.children);
-      // Reset value to null to prevent unwanted selection
-      this.value = null;
     });
     observer.observe(this.el, config);
   }


melaniebmn (approved):
LGTM

PR #1098, opened 2026-01-12
remove(gcds-header): remove signature variant in header

ethanWallace:
# Summary | Résumé

As part of the component API alignment work leading to stable v1.0.0 release, remove `signature-variant` property from the `gcds-header`.

## Design rationale

Using the `white` variant of the `gcds-signature` component within the `gcds-header` component renders the `gcds-signature` in white while leaving the rest of the built in elements in their normal colour scheme. This creates a disconnect between the signature and the rest of the components. If a developer needs to use a `white` signature, the signature can still be passed in the `signature` slot.


melaniebmn (changes requested):
Looks good, I think we still need to remove the toggle in Storybook as well though.

PR #1120, opened 2026-02-03
feat(gcds-header): add sign in button

Garneauma:
# Summary | Résumé

Adding option to include a Sign in link in the main header. The link has a customizable href and label.

The sign in link should appear on the right, under the header border. If there is a Themes and Topics menu, the link should be to the right end of it. If no Themes and Topics menu, the link should next to the breadcrumbs.

# Notes

I was not sure how to handle the CSS changes considering you seem to use CSS variables for everything but those variables are located in another project. Let me know how I should proceed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28101,6 +28101,7 @@
       "dev": true,
       "license": "MIT",
       "optional": true,
+      "peer": true,
       "dependencies": {
         "sass": "1.97.3"
       }
@@ -28118,6 +28119,7 @@
       "os": [
         "android"
       ],
+      "peer": true,
       "engines": {
         "node": ">=14.0.0"
       }
@@ -28135,6 +28137,7 @@
       "os": [
         "android"
       ],
+      "peer": true,
       "engines": {
         "node": ">=14.0.0"
       }
@@ -28152,6 +28155,7 @@
       "os": [
         "android"
       ],
+      "peer": true,
       "engines": {
         "node": ">=14.0.0"
       }
@@ -28169,6 +28173,7 @@
       "os": [
         "android"
       ],
+      "peer": true,
       "engines": {
         "node": ">=14.0.0"
       }
@@ -28186,6 +28191,7 @@
       "os": [
         "darwin"
       ],
+      "peer": true,
       "engines": {
         "node": ">=14.0.0"
       }
@@ -28203,6 +28209,7 @@
       "os": [
         "darwin"
       ],
+      "peer": true,
       "engines": {
         "node": ">=14.0.0"
       }
@@ -28220,6 +28227,7 @@
       "os": [
         "linux"
       ],
+      "peer": true,
       "engines": {
         "node": ">=14.0.0"
       }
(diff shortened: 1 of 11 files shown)


daine:
@Garneauma Thanks again for putting together this PR, amazing work on this! Thanks to Chelsey for providing answers and helpful examples. We've taken some time to review and analyze the current proposed implementation along with the feature request (#1139). The addition of a sign in link addresses a clear and specific need for users. We’re recommending some changes to the PR to better align with GCDS component patterns and standards.

We propose removing the properties `signInHref` and `signInLabel` and adding a new slot named `account` instead.

This would look like:
```
<gcds-button slot="account">
 Sign In
</gcds-button>
```

A few conditions apply:

- The `account` slot is optional and can accept any element. If a `gcds-button` is used in this space, it can be styled with a 0 border radius
- If both the Theme and Topic menu and `account` slot are present, they should appear on the same row, with the `account` slot aligned to the right.
- If both the breadcrumbs and `account` slot are present, they should appear on the same row, with the `account` slot aligned to the right
- If all three (Theme and topic menu, breadcrumbs, `account` slot) are present, the `account` slot should be aligned to the right, on the same row as the Theme and topic menu
- If only the `account` slot is present, it should appear on the right, directly below the divider line.

This change will help us to address your immediate need while fulfilling our aim to provide the most relevant and flexible solution for teams and departments who use GC Design System or will be adopting it in the near term. 
We also believe this provides a more flexible option that would help systematize sign in and account patterns by providing a consistent location and placement across more Government of Canada services. It  provides future opportunities to scale the sign in button and may also help to provide more research and data on account and sign in operations for future iterations. 
Thank you for your contribution and continued collaboration.

Garneauma:
@daine How would you go about adding the styles to the `<a>` tag? It seems the tag is not accessible through `gcds-header.css` when the tag is a slot. I've tried a few different ways but none have worked. Only solution I could figure out was to add styles inside `gcds-button.css`. Thanks!

daine:
> @daine How would you go about adding the styles to the `<a>` tag? It seems the tag is not accessible through `gcds-header.css` when the tag is a slot. I've tried a few different ways but none have worked. Only solution I could figure out was to add styles inside `gcds-button.css`. Thanks!

There's a few different ways to style things, if we're styling the slot, it's best to target the use of `<gcds-button` itself. I'll reach out to you via email with some specific code examples.

Garneauma:
@daine The component has been updated to now use a slot instead of leveraging attributes on the tag. Let me know if there's anything I should change. Thanks!

daine:
@Garneauma 
This is looking great! I had a quick look at the code, but before we do a full in-depth review, would you mind fixing the failing tests? Thank you!

Garneauma:
@daine I have fixed the failing tests. I have also modified a bit the logic inside the gcds-header rendering.

ethanWallace (approved):
LGTM

nmakuch (approved):
Was having some issues viewing this on my machine earlier but I'm able to view now and it LGTM as well 🫡👌

daine (changes requested):
LGTM but will review one last time as a group tomorrow

daine (approved):
@Garneauma looks great! Thanks for your patience on this.

PR #693, opened 2024-11-29
chore: Upgrade eslint and migrate to new flat config

radmorecameron:
# 📝 Summary | Résumé

This PR updates eslint to the latest version (flat config format). I opted to make a lot of things warnings for now. Most of the issues picked up by the linter can be fixed by running the `lint-and-fix` command.

I also added a lint stage to the test pipeline so the linter can be ran against future PRs to make sure there are no issues

## 🧩 Related Issues | Cartes liées

- Issue #1221 (my comment)

## 🧪 Test instructions | Instructions pour tester la modification

- [ ] Checkout this branch
- [ ] run `npm install`
- [ ] run `npm run lint`
- warnings should show but no errors

## ✍️ Author checklist | Liste de vérification de l'auteur

**Choose one:**
- [x] This PR does not introduce changes that need to be published on NPM (use `chore:`, `docs:`, `ci:`)

**Breaking changes flag:**
- [x] This PR does not break existing functionality. I have completely tested the functionality of these changes. 
- [x] This PR does not introduce any changes to component names, properties, values, or behaviour.

**Ready for review: (all items must be checked)**
- [x] Test instructions are clear and reproducible.

## 🧐 Reviewer checklist | Liste de vérification du réviseur

**Developer checklist**

For PRs that are complex, in lieu of a simple approval or an "LGTM" ✅, paste the following info with your approval:
- [ ] I have tested the changes and functionality using the test instructions.
- [ ] I have confirmed test coverage is adequate.
- [ ] I have reviewed the code for clarity, maintainability and potential issues.

**Design checklist (if needed)**

For designers, include the following info with your approval:

- [ ] I have tested the changes and functionality using the test instructions.
- [ ] The changes meet design expectations and align with the design system.
- [ ] Any design inconsistencies have been raised on slack or tracked via an issue.

**Content checklist (if needed)**

For content, include the following info with your approval:

- [ ] I understand the context and intent of the content changes.
- [ ] I have reviewed all content for clarity, readability, and tone.
- [ ] I have reviewed English and French content for accuracy and parity.

## ⚠️ Impact/Risks | Risques

_Optional: Use this section to highlight any potential implcations, risks or important notes for reviewers or maintainers, i.e. breaking changes, performance implications, dependency updates, etc._
_Highlight any deprecation notices here_

PR #1126, opened 2026-02-04
Playwright - remove try catch and fix lint errors

radmorecameron:
# Summary | Résumé

Removes try catch from playwright accessibility tests. If accessibility fails, then you'd want the test to fail too.
Makes a lot of lint fixes for #693 to be merged more smoothely (remove redundant awaits, use import instead of require)

Run the playwright tests 😄


radmorecameron:
GitHub didn't let me edit the title for the PR :upside_down_face:

PR #1332, opened 2026-07-13
fix(gcds-table): fix results to match filtered results and not total rows

daine:
This PR was opened to apply the suggestions from the code quality tool (before we turn it off)

_This PR applies 3/4 suggestions from code quality [AI findings](https://github.com/cds-snc/gcds-components/security/quality/ai-findings). 1 suggestion was skipped to avoid creating conflicts._

AI findings:
> In the filtered pagination branch, the `{end}` replacement uses `totalRows` as the cap in `Math.min`, but it should use `filteredRows` instead. When a filter is active, the total number of rows the user can page through is `filteredRows`, not `totalRows`, so the displayed "end" value can exceed the actual number of filtered results.

Note: This fix still needs to be tested and verified

### DIFF
--- a/packages/web/src/components/gcds-table/utils/render-helpers.tsx
+++ b/packages/web/src/components/gcds-table/utils/render-helpers.tsx
@@ -120,7 +120,7 @@ const renderTableStatus = (
       )
       .replace('{total}', totalRows);
 
-    // No rows avalable
+    // No rows available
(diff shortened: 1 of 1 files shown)


ethanWallace:
I get the logic what it is saying but with how it is now it does show the proper number already

Issue #1004, opened 2025-11-06
bug: Properties don't support rich text


s-laugh:
### Prerequisites | Prérequis

- [x] I have read the [Contributing Guidelines](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md). | J’ai lu les [Lignes directrices sur les contributions](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md). | J’accepte de me conformer au [Code de conduite](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/cds-snc/gcds-components/issues) that already report this problem, without success. | J’ai vérifié parmi [les tickets (issues) existants](https://github.com/cds-snc/gcds-components/issues) que ce sujet n’avait pas déjà été soumis.

### GC Design System Components Package and Version | Paquet et version des composants de Système de design GC

gcds-components@latest

### Current Behavior | Comportement observé

Currently every component (that I've try with so far, eg. card, input) that accepts text in a property (not a slot) that will be displayed to the user only accepts plain text. This means I can't provide some critical elements specifically an `abbr` tag around acronyms, that would be redundant and a nuisance to development and users to spell out every time in every place.

### Expected Behavior | Comportement attendu

Have properties accept rich text, or move them to slots.

### System Info | Information sur le système

```shell

```

### Steps to Reproduce | Étapes pour reproduire le bogue

```html
<gcds-card card-title="Statement of Account (<abbr title='British Columbia Training & Education Savings Grant'>BCTESG</abbr>) - Beneficiary or Parent"
           card-title-tag="h2"
           href="#"
           badge="<abbr title='Content Managment Solution'>CMS</abbr>"
           description="To produce a <abbr title='British Columbia Training & Education Savings Grant'>BCTESG</abbr> statement of account for a beneficiary or parent.">
</gcds-card>
```

<img width="405" height="505" alt="Image" src="https://github.com/user-attachments/assets/6020ad4e-6053-40a4-bbef-3f7586979a5a" />

In this case, the badge doesn't even render.

### Additional Information | Informations supplémentaires

[...]

Issue #1375, opened 2026-08-06
docs: no documented way to run the visual regression suite locally


ken-talltree-io:
### Context

I'm working on a fix for #1244 and wanted to confirm it caused no visual change before opening a PR. I couldn't find a supported way to do that locally.

### Concern

`CONTRIBUTING.md` documents `npm run build` and the unit/e2e tests, but says nothing about the visual regression suite. The only guidance I found is a comment inside `gcds-card.visual.ts`:

> To update baselines after an intentional visual change, trigger the "Update visual snapshots" workflow in GitHub Actions on your branch.

That covers *updating* baselines, not *running* the suite to check whether your change moved anything.

When I did run it locally, reproducing the CI steps from `run-tests.yml` (`npm ci` → `npx playwright install --with-deps chromium` → root `npm run build` → `npm run test:visual`), several components failed on an unmodified checkout:

```
✘ gcds-breadcrumbs › default
✘ gcds-breadcrumbs-item › default
✘ gcds-alert › default
✘ gcds-button › disabled
```

These look like platform rendering differences — baselines are generated on CI's Linux runners, I'm on macOS/arm64, and `snapshotPathTemplate` has no platform segment, so there's one baseline set shared across platforms. `tests/base.visual.ts` goes to real trouble to make rendering deterministic (intercepting Google Fonts and the design-system CDN so text metrics are stable), which suggests platform-stability is already a known concern.

The practical effect is that a contributor on a non-Linux machine can't use the suite to answer "did my change alter any rendering?", because the signal is mixed in with pre-existing platform noise.

### Suggested change

Any one of these would help, roughly in order of effort:

1. A short **"Running the visual regression tests"** section in `CONTRIBUTING.md` — the command, and an explicit note that local failures on other components are expected off-Linux and that CI is the source of truth.
2. Add a platform segment to `snapshotPathTemplate` so baselines can coexist per-platform.
3. Document a container-based run (matching the CI image) so local results are comparable.

I'd be glad to open a PR for option 1 if that's useful — it's the cheapest and it's what I'd have wanted today.

### Rationale

The visual suite is the only thing that can confirm a component change is pixel-safe. Right now an outside contributor either skips that check or reports results that aren't meaningful. For a change touching rendering, that gap is the difference between a PR a reviewer can trust and one they have to re-verify themselves.

### Environment

- macOS 26.3.1, arm64
- Node 22.15.0 (matching CI's version)
[...]

PR #1380, opened 2026-08-10
feat(gcds-card): text being skipped by Android and iOS read-aloud

### LINKED ISSUE #1244: bug: gcds-card `card-title` and `description` are inaccessible to textContent-based TTS tools (Android "Read Aloud", iOS "Listen to Page"
### Prerequisites | Prérequis

- [x] I have read the [Contributing Guidelines](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md). | J’ai lu les [Lignes directrices sur les contributions](https://github.com/cds-snc/gcds-components/blob/main/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md). | J’accepte de me conformer au [Code de conduite](https://github.com/cds-snc/gcds-components/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/cds-snc/gcds-components/issues) that already report this problem, without success. | J’ai vérifié parmi [les tickets (issues) existants](https://github.com/cds-snc/gcds-components/issues) que ce sujet n’avait pas déjà été soumis.

### GC Design System Components Package and Version | Paquet et version des composants de Système de design GC

latest

### Current Behavior | Comportement observé

The `gcds-card` component renders its `card-title` and `description` attributes entirely inside the Shadow DOM with no corresponding slot or light DOM equivalent. This makes both properties completely invisible to `textContent`-based text-to-speech tools, including Android Chrome's "Read Aloud" / "Read this page" and iOS Safari's "Listen to Page" feature. These tools are widely used on mobile by users with visual impairments, low literacy, or cognitive disabilities.

**Root cause**

When `card-title` and `description` are provided as attributes, the component renders them directly into the Shadow DOM:

```html
<!-- Shadow DOM — invisible to textContent and TTS -->
<div class="gcds-card">
  <h3 class="gcds-card__title">
    <gcds-link>Plan your departure</gcds-link>
  </h3>
  <div class="gcds-card__description">
    <gcds-text>Timeline, medical exams...</gcds-text>
  </div>
</div>
```

The light DOM of the element is empty. There is no slot for `card-title` at any point. The `description` slot only appears when the `description` attribute is omitted — but even then, only the description is slotted; the title remains permanently in the Shadow DOM.

**Impact**

This affects any Government of Canada page using `gcds-card` for navigation. Android "Read Aloud" stops reading as soon as it encounters the first empty `gcds-card`, cutting off all content that follows — including content in other components further down the page (headings, nav links, etc.) that would otherwise be fully readable.

WCAG 2.1 AA compliance via the ARIA accessibility tree (screen readers like TalkBack, VoiceOver) is not affected. This is specifically a `textContent` / DOM text extraction issue affecting browser-native TTS features that do not pierce the Shadow DOM.

### Expected Behavior | Comportement attendu

`document.querySelector('gcds-card').textContent` should return the card title and description text so that all DOM text extraction tools can read it.

### System Info | Information sur le système

```shell
- `@gcds-core/components@1.2.0` (CDN)
- Chrome 124 on Android 14 ("Read this page" reads nothing after the first heading that precedes the card grid)
- Tested on a live Government of Canada page using the component as documented
```

### Steps to Reproduce | Étapes pour reproduire le bogue

1. Use `gcds-card` with `card-title` and `description` attributes as documented:

```html
<gcds-card
  card-title="Plan your departure"
  card-title-tag="h3"
  href="before-avant/index.aspx?lang=eng"
  description="Timeline, medical exams, financial planning.">
</gcds-card>
```

2. In the browser console, run:

```js
document.querySelector('gcds-card').textContent.trim()
// Returns: ""  ← empty string
```

3. On an Android device, open the page in Chrome and activate "Read Aloud" / "Read this page." The card is entirely skipped — no title, no description.

### Code Reproduction URL | URL de reproduction du code

_No response_

### Additional Information | Informations supplémentaires

**Suggested fix**

Expose `card-title` in the light DOM by adding a named slot, and ensure the title text is present in the light DOM when the attribute is used. One approach:

```html
<!-- Shadow DOM -->
<div class="gcds-card">
  <h3 class="gcds-card__title">
    <gcds-link><slot name="card-title"></slot></gcds-link>
  </h3>
  <div class="gcds-card__description">
    <slot></slot>
  </div>
</div>
```

```html
<!-- Light DOM (auto-populated from attributes by the component) -->
<gcds-card card-title="Plan your departure" ...>
  <span slot="card-title">Plan your departure</span>
  <p>Timeline, medical exams, financial planning.</p>
</gcds-card>
```

Alternatively, the component could internally mirror attribute values into hidden light DOM text nodes (using `aria-hidden="true"` on the shadow content and readable text in the light DOM), or inject a visually-hidden `<span>` in the light DOM containing the title and description text.

**Workaround**

Until this is resolved, pages can replace `gcds-card` with plain HTML that visually matches the component while keeping all text in the light DOM. See the CSS required to replicate the card's appearance without the component in the meantime.

ken-talltree-io:
Closes #1244.

## What was happening

`gcds-card` renders `card-title` and `description` entirely inside its shadow root, so `document.querySelector('gcds-card').textContent` returns `""`.

That makes the card invisible to DOM-text extraction — including the browser-native read-aloud features on Android Chrome ("Read Aloud") and iOS Safari ("Listen to Page"). As @patrickpilon reported, those tools **stop at the first unreadable card**, so everything below it on the page goes unread too, not just the card.

The ARIA tree is unaffected: TalkBack and VoiceOver behave the same before and after, and automated checkers pass it clean either way. This is only visible to `textContent`-based extraction, which is why it hasn't shown up in tooling.

## What changed

`card-title` and `description` now render through named slots, filled from a mirror node the component maintains in the light DOM:

```html
<gcds-card card-title="Plan your departure" href="…" description="Timeline, medical exams.">
  <span slot="card-title" data-gcds-text-mirror>Plan your departure</span>
  <span slot="card-description" data-gcds-text-mirror>Timeline, medical exams.</span>
</gcds-card>
```

The slots carry the prop as **fallback content**, so if a mirror is ever absent (SSR, pre-hydration) the rendering is exactly what it is today. Visual output is unchanged — the styling wrappers stay in the shadow root and only the text is slotted.

Cards that fail validation render nothing, and are left without mirrors, so a broken card contributes no stray text.

## The part worth reviewing

`renderDescription` branched on light-DOM emptiness:

```ts
if (this.el.innerHTML.trim() != '') { … <slot></slot> … }
else if (this.description) { … }
```

**Both fixes suggested in the issue inject light-DOM nodes, which flips that guard.** The unnamed slot then renders, the injected node carries a `slot` name so it isn't assigned to it, and the `description` attribute is never rendered — the description would silently vanish from every card using the attribute API. The fix captures `hasSlottedDescription` in `componentWillLoad` *before* any mirror is written, so the component's own injection can't flip it. `'still renders an attribute description once the title is mirrored'` guards it.

## Tests

21 in `gcds-card.spec.tsx` (539 across the web package's spec suite). Six new cases; **four fail with the component reverted to `main`**:

- exposes a title given as an attribute
- exposes a description given as an attribute
- does not duplicate a description supplied through the default slot
- tracks a title changed after first render

The other two — `still renders an attribute description once the title is mirrored` and `leaves no text behind for a card that fails validation` — pass on `main`. They guard against defects *this change* could introduce rather than proving the bug is fixed, which is a different claim and worth not conflating.

**14 existing `toEqualHtml` snapshots were updated**, since every one now includes the slot wrapper and the mirror span. That's the largest part of the diff and the hunk most worth scrutinising.

## Not verified: visual regression

I could not confirm this is visually inert, and I'd rather say so than imply otherwise.

Running the visual suite the way `run-tests.yml` does (`npm ci` → `playwright install` → root `npm run build` → `npm run test:visual`), several unrelated components fail on macOS (`gcds-breadcrumbs`, `gcds-alert`, `gcds-button`) — expected, since baselines are generated on CI's Linux and `snapshotPathTemplate` has no platform segment. But `gcds-card` **passed even with a deliberate 12px hot-pink border compiled into the served chunk**, producing a byte-identical screenshot. I couldn't determine why, so I haven't filed it as a bug — I've only raised the documentation gap in #1375.

The relevant risk: the card's reset contains `slot { display: initial; }`, so every `<slot>` here is a real inline box rather than being transparent. This change adds two. For inline text inside `gcds-link` that should render identically, and the existing default description slot already relies on the same rule — but "should" is doing work there, and CI's visual run is the thing that would actually confirm it.

## Notes

- No API change. `card-title` and `description` behave the same for consumers; the mirror nodes are an implementation detail marked with `data-gcds-text-mirror`.
- Two internal slots are documented in the component JSDoc, so `readme.md` and `specs/components.json` are regenerated by `stencil build --docs` and included here.

**AI disclosure:** written with Claude Code (Claude Opus). It explored the component, drafted the fix and tests, and ran the suites. I reviewed the result before opening this: that review is what caught the `renderDescription` guard interaction described above, which the issue's own suggested fixes would have tripped, and it is why the four genuinely-discriminating tests are distinguished above from the two that only guard against new defects. Accountability for the change is mine.

---

*Note on timing: I offered to take this on the issue on 6 August and haven't heard back, so I'm
opening it rather than letting the fix go stale — happy to close it if the team would rather keep
this one in-house.*

### DIFF
--- a/packages/web/specs/components.json
+++ b/packages/web/specs/components.json
@@ -1,5 +1,4 @@
 {
-  "timestamp": "2026-08-06T18:30:25",
   "compiler": {
     "name": "@stencil/core",
     "version": "4.43.4",
@@ -845,15 +844,11 @@
       "docsTags": [
         {
           "name": "slot",
-          "text": "default - Slot for the card description. Will overwrite the description prop if used."
+          "text": "title - Slot for the card title. Accepts rich text, so markup such as\n`<abbr>`, `<em>` or an icon can be used where the card-title prop cannot. Falls back to\nthe card-title prop, which is mirrored into the light DOM so the title text stays\nreadable by DOM-text extraction tools."
         },
         {
           "name": "slot",
-          "text": "card-title - Internal. Filled automatically from the card-title prop so the title\ntext exists in the light DOM and stays readable by DOM-text extraction tools."
-        },
-        {
-          "name": "slot",
-          "text": "card-description - Internal. Filled automatically from the description prop for the\nsame reason. Not used when the default slot is populated."
+          "text": "default - Slot for the card description. Accepts rich text and overwrites the\ndescription prop if used. The description prop is mirrored into this slot for the same\nDOM-text reason when nothing is slotted."
         }
       ],
       "usage": {},
@@ -1128,16 +1123,12 @@
       "styles": [],
       "slots": [
         {
-          "name": "card-description",
-          "docs": "Internal. Filled automatically from the description prop for the\nsame reason. Not used when the default slot is populated."
-        },
-        {
-          "name": "card-title",
-          "docs": "Internal. Filled automatically from the card-title prop so the title\ntext exists in the light DOM and stays readable by DOM-text extraction tools."
-        },
-        {
           "name": "default",
-          "docs": "Slot for the card description. Will overwrite the description prop if used."
+          "docs": "Slot for the card description. Accepts rich text and overwrites the\ndescription prop if used. The description prop is mirrored into this slot for the same\nDOM-text reason when nothing is slotted."
+        },
+        {
+          "name": "title",
+          "docs": "Slot for the card title. Accepts rich text, so markup such as\n`<abbr>`, `<em>` or an icon can be used where the card-title prop cannot. Falls back to\nthe card-title prop, which is mirrored into the light DOM so the title text stays\nreadable by DOM-text extraction tools."
         }
       ],
       "parts": [],
@@ -4478,7 +4469,7 @@
             "references": {
               "ContentValues": {
                 "location": "local",
-                "path": "/Users/kenbritton/src/oss/gcds-components/packages/web/src/components/gcds-grid/gcds-grid.tsx",
+                "path": "/web/src/components/gcds-grid/gcds-grid.tsx",
                 "id": "src/components/gcds-grid/gcds-grid.tsx::ContentValues"
               }
             }
@@ -4780,7 +4771,7 @@
             "references": {
               "GridGapValues": {
                 "location": "local",
-                "path": "/Users/kenbritton/src/oss/gcds-components/packages/web/src/components/gcds-grid/gcds-grid.tsx",
+                "path": "/web/src/components/gcds-grid/gcds-grid.tsx",
                 "id": "src/components/gcds-grid/gcds-grid.tsx::GridGapValues"
               }
             }
@@ -4876,7 +4867,7 @@
             "references": {
               "GridGapValues": {
                 "location": "local",
-                "path": "/Users/kenbritton/src/oss/gcds-components/packages/web/src/components/gcds-grid/gcds-grid.tsx",
+                "path": "/web/src/components/gcds-grid/gcds-grid.tsx",
                 "id": "src/components/gcds-grid/gcds-grid.tsx::GridGapValues"
               }
             }
@@ -4966,7 +4957,7 @@
             "references": {
               "GridGapValues": {
                 "location": "local",
-                "path": "/Users/kenbritton/src/oss/gcds-components/packages/web/src/components/gcds-grid/gcds-grid.tsx",
+                "path": "/web/src/components/gcds-grid/gcds-grid.tsx",
                 "id": "src/components/gcds-grid/gcds-grid.tsx::GridGapValues"
               }
             }
@@ -5056,7 +5047,7 @@
             "references": {
               "ContentValues": {
                 "location": "local",
-                "path": "/Users/kenbritton/src/oss/gcds-components/packages/web/src/components/gcds-grid/gcds-grid.tsx",
+                "path": "/web/src/components/gcds-grid/gcds-grid.tsx",
                 "id": "src/components/gcds-grid/gcds-grid.tsx::ContentValues"
               }
             }
@@ -5146,7 +5137,7 @@
             "references": {
               "ContentValues": {
                 "location": "local",
-                "path": "/Users/kenbritton/src/oss/gcds-components/packages/web/src/components/gcds-grid/gcds-grid.tsx",
+                "path": "/web/src/components/gcds-grid/gcds-grid.tsx",
                 "id": "src/components/gcds-grid/gcds-grid.tsx::ContentValues"
               }
             }
@@ -6624,7 +6615,7 @@
             "references": {
               "IconNames": {
                 "location": "local",
-                "path": "/Users/kenbritton/src/oss/gcds-components/packages/web/src/components/gcds-icon/gcds-icon.tsx",
+                "path": "/web/src/components/gcds-icon/gcds-icon.tsx",
                 "id": "src/components/gcds-icon/gcds-icon.tsx::IconNames"
               }
             }
--- a/packages/web/specs/components.json
+++ b/packages/web/specs/components.json
@@ -1,4 +1,5 @@
 {
+  "timestamp": "2026-08-06T18:30:25",
   "compiler": {
     "name": "@stencil/core",
     "version": "4.43.4",
@@ -845,6 +846,14 @@
         {
           "name": "slot",
           "text": "default - Slot for the card description. Will overwrite the description prop if used."
+        },
+        {
+          "name": "slot",
+          "text": "card-title - Internal. Filled automatically from the card-title prop so the title\ntext exists in the light DOM and stays readable by DOM-text extraction tools."
+        },
+        {
+          "name": "slot",
+          "text": "card-description - Internal. Filled automatically from the description prop for the\nsame reason. Not used when the default slot is populated."
         }
       ],
       "usage": {},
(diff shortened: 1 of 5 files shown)


ethanWallace:
Hey @ken-talltree-io thanks for the contribution. I am hoping to take a look at this pull request later this week.

ken-talltree-io:
Hey @ethanWallace — following up on this. I found and fixed an E2E regression the original fix introduced (`gcds-card.e2e.ts` was comparing `.innerText()` against `.textContent()` and only happened to agree pre-fix — full root-cause and verification in the commit), pushed at `01cfc57b` on 08-13. That re-triggered the first-time-contributor gate and it's been sitting at `action_required` since, so CI hasn't actually run on that head yet.

No rush on the review itself — just flagging that someone with access needs to approve the workflow run before there's a green (or red) signal to look at. Happy to answer anything in the meantime.

ethanWallace:
Hey @ken-talltree-io I was discussing this PR with the other devs and we are hoping you could do a slight refactor. We discussed having two new slots that are internal only may cause confusion to developers looking at the more technical documentation of the component and we see this as a great opportunity to solve https://github.com/cds-snc/gcds-components/issues/1004. 

Requested changes:

1. Change the `card-title` slot to just `title`.
2. Remove `card-description` slot and make the sync functions work with the `default` slot.
3. Remove the `internal` flags from the slot descriptions.

Thank you, we really appreciate the effort to help solve this issue.

daine (commented):
Thanks for this @ken-talltree-io ! Just adding one small addition to @ethanWallace's comment above

packages/web/specs/components.json
```diff
@@ -4461,7 +4478,7 @@
             "references": {
               "ContentValues": {
                 "location": "local",
-                "path": "/web/src/components/gcds-grid/gcds-grid.tsx",
+                "path": "/Users/kenbritton/src/oss/gcds-components/packages/web/src/components/gcds-grid/gcds-grid.tsx",
```
daine: We have a script that's supposed to clean up the paths, but for some reason it's picking up yours. Would you mind doing the cleanup so your personal directory path isn't on this file? Thank you!

```suggestion
                "path": "/web/src/components/gcds-grid/gcds-grid.tsx",
```

ken-talltree-io:
Thanks both — this is a better shape than what I had, and the rich-text case is a much better reason for the slots to exist than the one I gave them. All three changes are in `87fa876d`.

- `card-title` slot → `title`.
- `card-description` is gone. The description prop mirrors into the `default` slot instead, so there's one description slot rather than two.
- `internal` flags removed from both descriptions.

## One change that wasn't a rename

Making the slots public exposed a bug in my own code, so flagging it rather than burying it in the diff.

Author content was detected with `this.el.innerHTML.trim() != ''`. That was fine while the slots were internal, because the only light-DOM content a card could have *was* the description. Once `title` is public it breaks: a card that slots only a title reads as having slotted description content, and the `description` prop is dropped silently.

```html
<gcds-card card-title="Passports" href="#p" description="Renew online.">
  <span slot="title">Passports</span>
</gcds-card>
<!-- before: "Renew online." never renders -->
```

So detection is now per-slot — `hasSlottedContent()` looks at child nodes for the relevant slot assignment and ignores the component's own mirrors. A slotted title also suppresses the title mirror, so an author's markup isn't rendered alongside the prop text.

## @daine — the paths

Done, and there were **7 of them**, not just the one you flagged: six on `gcds-grid.tsx` entries (4481, 4783, 4879, 4969, 5059, 5149) and one on `gcds-icon.tsx` (6627).

I think I know why the script "picked up" mine. `postbuild` is an npm lifecycle hook on `build`, so `npm run build` sanitizes correctly — I re-ran it just now and it printed `✅ Paths sanitized` with zero absolute paths left. But invoking `stencil build` directly, or anything that skips npm lifecycle scripts, writes `components.json` without ever running `postbuild.js`. That's almost certainly what I did. Nothing wrong with your script — it just isn't reachable on that path. Happy to open a separate PR making it defensive (a `prepublish`-time check, or sanitizing inside the docs output target) if that's useful.

## Should this close #1004 as well?

Happy to add `Closes #1004` to the description if you agree it covers it. From the thread it looks like it does — `<abbr>` for @s-laugh's original report, icons for @radmorecameron, and `<em>` for @tobiG1t's bilingual case all work through the `title` slot now. Tell me if you'd rather keep #1004 open for the other components with the same property-vs-slot gap, and I'll leave it.

## Verification

542 unit tests passing, up from 539 — three added for the slot behaviour. The one that matters is the description case above; it isn't padding, and restoring the old `innerHTML` check gives exactly 1 failure, that one. Card e2e is 7/7 including the Keyboard focus case. Full unit suite and the rest of the repo are unchanged against a stashed baseline.

ethanWallace (approved):
LGTM. Thank you @ken-talltree-io for the contribution. Let's keep https://github.com/cds-snc/gcds-components/issues/1004 open so we can have a work item to expand this to other components. I am also going to change the PR's title to feat to have this marked as a feature update since a new `title` slot has been added.

- [X] I have tested the changes and functionality using the test instructions.
- [X] I have confirmed test coverage is adequate.
- [X] I have reviewed the code for clarity, maintainability and potential issues.

PR #1384, opened 2026-08-12
docs: document how to run the visual regression suite locally

### LINKED ISSUE #1375: docs: no documented way to run the visual regression suite locally

ken-talltree-io:
Closes #1375.

@daine — thanks for the quick reply on the issue, and for explaining the reasoning. This is the `CONTRIBUTING.md` update you asked for.

Your framing changed what I wrote. I had assumed the missing local workflow was an oversight; you said CI is deliberately the source of truth because it's the cheapest path and you see failures close to release. So the section documents the local run as a **secondary aid** and says plainly that CI decides — rather than implying the local result is authoritative.

### What's here

A `Running the visual regression tests` section, in both official languages, placed after `Compiling all packages` in each.

It documents the four steps CI already runs, and the three things that aren't discoverable without reading the config:

- **The dev server starts itself.** `@stencil/playwright`'s `createConfig` supplies `stencil build --dev --watch --serve --no-open`, so there's no second terminal. Nothing said this, and it's the first thing a contributor would get wrong.
- **`npm run test:visual` doesn't build.** The `pretest` hook in `packages/web` only fires for `test`, not `test:visual`, so a stale build silently produces meaningless results.
- **Locally generated baselines must not be committed**, and the `Update visual snapshots` workflow **can't be triggered from a fork** — it's `workflow_dispatch` (write access) and checks out `inputs.branch` from this repo, not the contributor's. So an outside contributor's only route is to ask a reviewer. That seemed worth stating outright rather than leaving someone to discover it after an intentional visual change.

On the off-Linux failures, I used your own measurement rather than mine: `playwright.visual.config.ts` already records that a local macOS render differs by ~1% against the 0.5% `maxDiffPixelRatio`. That's a clearer explanation than the one in my issue, and it's the number a contributor can go verify.

### Scope

This is **option 1** of the three I listed in #1375. I've deliberately not pursued the other two — a platform segment in `snapshotPathTemplate`, or a documented container run — because both make the local result authoritative, which is the opposite of how you described the design. Happy to open either separately if you'd want it, but I'd rather not assume.

### Verification

Docs-only change; no code, no tests, no CI behaviour touched. I ran what the section tells people to run, on a clean checkout of `main` (`1725b29`), macOS 26.3.1 / arm64:

- Both invocations work as written — root `npm run test:visual` and `npm run test:visual` from `packages/web`.
- The dev server does start and stop on its own.
- 10 tests fail on components untouched by this branch (6 × `gcds-alert`, `gcds-breadcrumbs`, `gcds-breadcrumbs-item`, `gcds-button` × 2), every one reporting `ratio 0.01` against the 0.005 tolerance — which is exactly the ~1% the config comment predicts. That's the noise the new section warns about, reproduced.

### One unrelated thing I noticed

`run-tests.yml:88` has the artifact upload at `if: always() # TEMP debug`, with the intended `if: failure() && steps.visual.outcome == 'failure'` commented out on line 87 — so the visual report uploads on every run, including green ones. I've left it alone since it's outside this PR. Say the word and I'll send a one-line fix, or ignore it if it's intentional for now.


daine (dismissed):
Thank you for adding this information on our `CONTRIBUTING.md` docs, it truly is much appreciated.

It looks good to me!

CONTRIBUTING.md
```diff
@@ -149,6 +170,27 @@ Vous pouvez tester localement les paquets Angular, Vue et React puisque ce réf
...
+
+**Ne versionnez pas d’images de référence générées localement.** `npm run test:visual:update` remplace les images de référence par des rendus provenant de votre machine, ce qui brise la suite pour tout le monde si ces rendus ne proviennent pas de Linux. Pour mettre à jour les images de référence après un changement visuel intentionnel, exécutez le flux de travail **Update visual snapshots** dans GitHub Actions sur votre branche. Si vous travaillez à partir d’une duplication (fork), ce flux de travail ne peut pas accéder à votre branche : demandez à une personne réviseure de l’exécuter pour vous.
```
EliseKa:
```suggestion
**Ne versez pas d’images de référence générées localement.** `npm run test:visual:update` remplace les images de référence par des rendus provenant de votre machine, ce qui brise la suite pour tout le monde si ces rendus ne proviennent pas de Linux. Pour mettre à jour les images de référence après un changement visuel intentionnel, exécutez le flux de travail **Update visual snapshots** dans GitHub Actions sur votre branche. Si vous travaillez à partir d’une duplication (fork), ce flux de travail ne peut pas récupérer votre branche : demandez à une personne réviseure de la demande de tirage de l’exécuter pour vous.
```

CONTRIBUTING.md
```diff
@@ -149,6 +170,27 @@ Vous pouvez tester localement les paquets Angular, Vue et React puisque ce réf
...
+
+Lorsque la vérification visuelle échoue dans une demande de tirage, téléchargez l’artéfact `visual-regression-report-<numéro d’exécution>` de l’exécution et ouvrez `index.html`. Il présente côte à côte l’image attendue, l’image obtenue et leur différence pour chaque capture en échec.
```
EliseKa:
```suggestion
Lorsque la vérification visuelle échoue dans une demande de tirage, téléchargez l’artéfact `visual-regression-report-<numéro d’exécution>` généré par le flux de travail et ouvrez `index.html`. Pour chaque capture d'écran ayant échoué, ce fichier présente côte à côte l’image attendue, l’image obtenue et leur différence.
```

EliseKa (commented):
French looks good, recommending only a few edits :)

daine (approved):

ethanWallace (approved):
LGTM

